PR #3, opened 2023-01-05
Create initial version of library

litt3:
There are 2 main parts to this cryptography repo:
1. An interface between java and rust, to be able to access the crypto implementation in the [zk crypto BLS library](https://github.com/zkcrypto/bls12_381)
    - The rust side of the interface can be found in the `hedera-bls-rust-jni` module of this repository
    - The java side of the interface can be found in the file `BLS12381Bindings` in the `hedera-bls-api` module
    - The `LibraryLoader` class is responsible for loading the rust library in, and setting up the bindings
2. A set of java objects representing the BLS cryptographic primitives, in the `hedera-bls-api` module
    - these objects call into the java/rust interface to perform their function

For some context behind the cryptographic primitives implemented here and their supported operations, see [this section](https://www.notion.so/swirldslabs/BLS-9c9339ee83b441268fe11931b429e58f#956e7ad045f146ce93397137eede3ae9) of the BLS notion doc


lpetrovic05:
Can you give a bit more context, I'm not sure what I'm reviewing

litt3:
FYI - The `LibraryLoader` class is rather unpolished at the moment. I'm holding off putting a ton of time into that until we're further along in the build pipeline effort, since it seems possible that things could change about how we are loading the native library.

.idea/google-java-format.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
lpetrovic05: I think we should use the same setup we do for other repos. Best to talk to @nathanklick about this

lpetrovic05:
It doesn't seem like we should have jars in the repo

hedera-bls-api/src/main/java/com/hedera/platform/bls/BLS12381BilinearMap.java
```diff
@@ -0,0 +1,117 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.hedera.platform.bls;
+
+/** A bilinear map in the BLS 12-381 family of curves */
+public final class BLS12381BilinearMap implements BilinearMap {
+    /** {@inheritDoc} */
+    @Override
+    public Field getField() {
```
lpetrovic05: I don't think that methods that create new instances should start with `get`, its against the java convention

hedera-bls-api/src/main/java/com/hedera/platform/bls/BLS12381BilinearMap.java
```diff
@@ -0,0 +1,117 @@
...
+            final GroupElement keyElement2) {
+
+        if (signatureElement1 == null
+                || keyElement1 == null
+                || signatureElement2 == null
+                || keyElement2 == null) {
+            throw new IllegalArgumentException("all arguments must be valid");
+        }
+
+        return BLS12381Bindings.comparePairing(
```
lpetrovic05: blind casting, we should cast with `instanceof` and throw an appropriate exception if the type is not correct

hedera-bls-api/src/main/java/com/hedera/platform/bls/BLS12381BilinearMap.java
```diff
@@ -0,0 +1,117 @@
...
+            final GroupElement signatureElement1,
+            final GroupElement keyElement1,
+            final GroupElement signatureElement2,
+            final GroupElement keyElement2) {
+
+        if (signatureElement1 == null
+                || keyElement1 == null
+                || signatureElement2 == null
+                || keyElement2 == null) {
+            throw new IllegalArgumentException("all arguments must be valid");
```
lpetrovic05: not a very informative message, it should say that an argument must not be null

hedera-bls-api/src/main/java/com/hedera/platform/bls/BLS12381BilinearMap.java
```diff
@@ -0,0 +1,117 @@
...
+    public boolean equals(final Object o) {
+        if (this == o) {
+            return true;
+        }
+
+        if (o == null) {
+            return false;
+        }
+
+        return getClass() == o.getClass();
```
lpetrovic05: only comparing if its the same class? why?

hedera-bls-api/src/main/java/com/hedera/platform/bls/BLS12381Exception.java
```diff
@@ -0,0 +1,49 @@
...
+        errorCodeMap.put(4, "InputLength");
+    }
+
+    /**
+     * Constructor
+     *
+     * @param functionName the name of the library function where the error occurred
+     * @param errorCode the error code that was returned
+     */
+    public BLS12381Exception(final String functionName, final int errorCode) {
```
lpetrovic05: better to use an enum if the error codes are limited

hedera-bls-api/src/main/java/com/hedera/platform/bls/Field.java
```diff
@@ -0,0 +1,71 @@
...
+
+/** Interface representing a generic field */
+public interface Field {
+    /**
+     * Creates a new field element from a long
+     *
+     * @param inputLong the long to use to create the field element
+     * @return the new field element
+     */
+    FieldElement elementFromLong(final long inputLong);
```
lpetrovic05: wouldn't it make more sense to have these methods that generate fields in a separate interface? atm, this interface represents a field, as well as a field factory

litt3: These methods are actually generating field *elements*, as opposed to fields.

Essentially, the `Field` class is a `FieldElement` factory

.idea/google-java-format.xml
litt3: Nathan committed this file - I'm not sure if it was intended to be temporary, but I know he still has outstanding TODOs before this repo is ready for release. I'll make sure the formatting is considered before that :+1:

hedera-bls-api/src/main/java/com/hedera/platform/bls/BLS12381BilinearMap.java
litt3: @lpetrovic05 is the preferred style to have a separate check for each argument, so that the error message can specify *which* parameter was null? Or is it preferred to check them all in 1 block (like the current check), but with a better error message?

lpetrovic05: In platform we use a check per argument:
```
	/**
	 * Throw an {@link IllegalArgumentException} if the supplied argument is {@code null}.
	 *
	 * @param arg
	 * 		the argument checked
	 * @param argName
	 * 		the name of the argument
	 */
	public static <T> T throwArgNull(final T arg, final String argName) {
		if (arg == null) {
			throw new IllegalArgumentException(
					String.format(
							"The supplied argument '%s' cannot be null!",
							argName
					)
			);
		}
		return arg;
	}
```

hedera-bls-api/src/main/java/com/hedera/platform/bls/Field.java
lpetrovic05: ok, got it. can the javadoc of the class explain this?

litt3:
> It doesn't seem like we should have jars in the repo

There is only `gradle-wrapper.jar`, which also exists in platform. I'm assuming therefore that it's required? @nathanklick can you clarify?

hedera-bls-api/src/main/java/com/hedera/platform/bls/BLS12381BilinearMap.java
litt3: Added static member variables to hold these objects, instead of creating new instances each call

hedera-bls-api/src/main/java/com/hedera/platform/bls/BLS12381BilinearMap.java
litt3: Fixed

hedera-bls-api/src/main/java/com/hedera/platform/bls/Field.java
litt3: Added

hedera-bls-api/src/main/java/com/hedera/platform/bls/BLS12381Exception.java
litt3: The issue I see with this is that rust can't return a java enum - which means we would need one of the two following strategies to do an `integer to enum` conversion:

1. Create an `errorEnumFromInteger` function, which would have to be manually updated with each change to the enum
2. Create a static `ErrorCode` enum array, set at construction via `enum.values()`.
    - Doing this would require us to check error code index when constructing an exception object (which is fine, but a bit messier)

Let me know which of these options you prefer, or if you had something else in mind

hedera-bls-api/src/main/java/com/hedera/platform/bls/BLS12381BilinearMap.java
litt3: Since the only member variables of this class are static, it isn't possible that 2 valid instances will *not* be equal

.idea/google-java-format.xml
lpetrovic05: @nathanklick pinging you again

hedera-bls-api/src/main/java/com/hedera/platform/bls/BLS12381BilinearMap.java
lpetrovic05: In that case, do we even need an `equals` and `hashCode`?

lpetrovic05: Will we ever be using them?

hedera-bls-api/src/main/java/com/hedera/platform/bls/BLS12381Exception.java
lpetrovic05: lets leave it as is, i don't want to complicate things unnecessarily

hedera-bls-api/src/main/java/com/hedera/platform/bls/BLS12381Field.java
```diff
@@ -0,0 +1,128 @@
...
+    /** Length of a byte array representing a field element */
+    public static final int ELEMENT_BYTE_SIZE = 32;
+
+    /** {@inheritDoc} */
+    @Override
+    public FieldElement elementFromLong(final long inputLong) {
+        final byte[] output = new byte[ELEMENT_BYTE_SIZE];
+
+        final int errorCode;
+        if ((errorCode = BLS12381Bindings.newScalarFromLong(inputLong, output)) != 0) {
```
lpetrovic05: `0` should be a constant

hedera-bls-api/src/main/java/com/hedera/platform/bls/BLS12381Field.java
```diff
@@ -0,0 +1,128 @@
...
+    }
+
+    /** {@inheritDoc} */
+    @Override
+    public int getSeedSize() {
+        return SEED_SIZE;
+    }
+
+    @Override
+    public boolean equals(final Object o) {
```
lpetrovic05: if these `equals` & `hashCode` methods are not needed, better to remove them I think

hedera-bls-api/src/main/java/com/hedera/platform/bls/BLS12381BilinearMap.java
```diff
@@ -0,0 +1,129 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.hedera.platform.bls;
+
+/** A bilinear map in the BLS 12-381 family of curves */
+public final class BLS12381BilinearMap implements BilinearMap {
```
lpetrovic05: If every instance of this class will always be the same, we should make it a singleton

hedera-bls-api/src/main/java/com/hedera/platform/bls/LibraryLoader.java
```diff
@@ -0,0 +1,357 @@
...
+ * @see <a
+ *     href="http://adamheinrich.com/blog/2012/how-to-load-native-jni-library-from-jar">http://adamheinrich
+ *     .com/blog/2012/how-to-load-native-jni-library-from-jar</a>
+ * @see <a
+ *     href="https://github.com/adamheinrich/native-utils">https://github.com/adamheinrich/native-utils</a>
+ */
+public final class LibraryLoader {
+    private static final String LIBRARY_NAME = "libhedera_bls_jni";
+
+    private final Logger logger = LogManager.getLogger(LibraryLoader.class);
```
lpetrovic05: a logger should be static

hedera-bls-api/src/main/java/com/hedera/platform/bls/LibraryLoader.java
```diff
@@ -0,0 +1,357 @@
...
+
+    /**
+     * Loads library from the current JAR archive and registers the native methods of the provided
+     * class. The library will be loaded at most once.
+     *
+     * <p>The file from JAR is copied into system temporary directory and then loaded. The temporary
+     * file is deleted after exiting.
+     */
+    public void loadBundledLibrary(final Class<?> clazz) throws IOException {
```
lpetrovic05: I think all of these methods should be static

hedera-bls-api/src/main/java/com/hedera/platform/bls/LibraryLoader.java
```diff
@@ -0,0 +1,357 @@
...
+     *     one.
+     * @return The file or directory that was requested.
+     * @throws IOException Could not find requested file.
+     */
+    private File getFileFromFileSystem(final Path relativePath, final File outputDirectory)
+            throws IOException, URISyntaxException {
+
+        final URL url =
+                ResourceLoader.class.getResource(
+                        prefixStringWithSlashIfNotAlready(relativePath.toString()));
```
lpetrovic05: if it starts with a slash, then its an absolute path, right? then there is a method like `Path.toAbsolute()` that should sort it out. also, this would not work on windows

hedera-bls-api/src/main/java/com/hedera/platform/bls/LibraryLoader.java
```diff
@@ -0,0 +1,357 @@
...
+ * https://github.com/terl/lazysodium-java/blob/master/src/main/java/com/goterl/lazysodium/utils/LibraryLoader.java
+ * </a>
+ *
+ * @see <a
+ *     href="http://adamheinrich.com/blog/2012/how-to-load-native-jni-library-from-jar">http://adamheinrich
+ *     .com/blog/2012/how-to-load-native-jni-library-from-jar</a>
+ * @see <a
+ *     href="https://github.com/adamheinrich/native-utils">https://github.com/adamheinrich/native-utils</a>
+ */
+public final class LibraryLoader {
```
lpetrovic05: are there any unit tests for this class?

hedera-bls-api/src/main/java/com/hedera/platform/bls/LibraryLoader.java
```diff
@@ -0,0 +1,357 @@
...
+
+    /**
+     * Copies a file into a temporary directory regardless of if it is in a JAR or not.
+     *
+     * @param relativePath A relative path to a file or directory, relative to the resources folder.
+     * @return The file or directory you want to load.
+     * @throws IOException If at any point processing of the resource file fails.
+     * @throws URISyntaxException If resource file cannot be found
+     */
+    public File copyToTempDirectory(Path relativePath, Class<?> outsideClass)
```
lpetrovic05: should this be public?

hedera-bls-api/src/main/java/com/hedera/platform/bls/BLS12381Group1.java
```diff
@@ -0,0 +1,149 @@
...
+
+    /** {@inheritDoc} */
+    @Override
+    public GroupElement batchMultiply(final Collection<GroupElement> elements) {
+        if (elements.isEmpty()) {
+            throw new IllegalArgumentException("Empty collection is invalid");
+        }
+
+        if (elements.contains(null)) {
+            throw new IllegalArgumentException("invalid element in collection");
```
poulok:
```suggestion
            throw new IllegalArgumentException("invalid element (null) in collection");
```

hedera-bls-api/src/main/java/com/hedera/platform/bls/GroupElement.java
```diff
@@ -0,0 +1,84 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.hedera.platform.bls;
+
+/** Interface representing a cryptographic group element */
+public interface GroupElement {
```
poulok: It seems like it would be better to parameterize the `GroupElement` and `FieldElement` interfaces since the implementing classes do a lot of type checking.

`FieldElement<T extends FieldElement>` and make the interface methods accept `T`

`GroupElement<R extends GroupElement, T extends FieldElement>` and use `R` and `T` in the methods as appropriate.

But maybe you tried this and it made other interfaces too complicated?

litt3: It used to be implemented this way, but it got quite complicated, such that many common types were carrying around 4+ generics.

I personally think the type checking in the implementing classes was worth the overall reduction in complexity, though I probably could be persuaded otherwise

hedera-bls-api/src/main/java/com/hedera/platform/bls/LibraryLoadingException.java
```diff
@@ -0,0 +1,24 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.hedera.platform.bls;
+
+/** Indicates a failure to load the required library. */
+public class LibraryLoadingException extends RuntimeException {
+
+    public LibraryLoadingException(String message) {
```
poulok:
```suggestion
    public LibraryLoadingException(final String message) {
```

hedera-bls-api/src/main/java/com/hedera/platform/bls/Utils.java
```diff
@@ -0,0 +1,40 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.hedera.platform.bls;
+
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
+
+public class Utils {
```
poulok:
```suggestion
public final class Utils {
```

hedera-bls-api/src/main/java/com/hedera/platform/bls/GroupElement.java
poulok: good enough reason for me to leave it as is

hedera-bls-api/src/main/java/com/hedera/platform/bls/BLS12381Field.java
litt3: Removed

hedera-bls-api/src/main/java/com/hedera/platform/bls/BLS12381BilinearMap.java
litt3: Because I implemented the factory pattern using the java `ServiceLoader` class, it isn't possible to make this bilinear map a singleton (since the service loader requires a public constructor). I did, however, make the field and group classes into singletons.

If we end up deciding not to go the `ServiceLoader` route, `BLS12381BilinearMap` could be converted to a singleton at that time.

hedera-bls-api/src/main/java/com/hedera/platform/bls/BLS12381BilinearMap.java
litt3: You are correct - they are no longer needed. Removing.

hedera-bls-api/src/main/java/com/hedera/platform/bls/LibraryLoader.java
litt3: Done

hedera-bls-api/src/main/java/com/hedera/platform/bls/LibraryLoader.java
litt3: Done

hedera-bls-api/src/main/java/com/hedera/platform/bls/LibraryLoader.java
litt3: Switched to private

hedera-bls-api/src/main/java/com/hedera/platform/bls/LibraryLoadingException.java
litt3: Fixed

hedera-bls-api/src/main/java/com/hedera/platform/bls/BLS12381Field.java
litt3: Done

hedera-bls-api/src/main/java/com/hedera/platform/bls/LibraryLoader.java
litt3: The problem is that this is a relative path, so calling `toAbsolute()` changes the path into one that isn't usable

hedera-bls-api/src/main/java/com/hedera/platform/bls/LibraryLoader.java
litt3: Not yet. Since I'm [still not entirely sure](https://github.com/hashgraph/hedera-bls-cryptography/pull/3#issuecomment-1380556728) if this way of loading in the native library will stick around, I haven't been spending too much time getting it production ready

hedera-bls-api/src/main/java/com/hedera/platform/bls/BLS12381Group1.java
litt3: Done

hedera-bls-api/src/main/java/com/hedera/platform/bls/Utils.java
litt3: Done

lpetrovic05 (approved):

buildSrc/src/main/kotlin/com.hedera.platform.rust-library.gradle.kts
```diff
@@ -0,0 +1,2 @@
+
+
```
cody-littley: why this empty file?

hedera-bls-api/src/main/java/com/hedera/platform/bls/api/BLSLoader.java
```diff
@@ -0,0 +1,49 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.hedera.platform.bls.api;
+
+import java.util.ServiceLoader;
+import java.util.concurrent.atomic.AtomicBoolean;
+
+/** Loader for accessing an implementation of {@link BilinearMap} */
+public final class BLSLoader {
```
cody-littley, line 22: Although this strategy may at first seem elegant, I actually think it's a bit of a trap. We are pushing to eliminate this style of stateful static object in the platform code base, since it makes testing much harder. 

I don't think this PR needs to be blocked on this issue, but we should certainly have some design discussions as a team and figure out how we want to load / instantiate these libraries.

hedera-bls-api/src/main/java/com/hedera/platform/bls/api/BilinearMap.java
```diff
@@ -0,0 +1,73 @@
...
+            final GroupElement keyElement2);
+
+    /**
+     * Computes a pairing, and returns a string representing the result
+     *
+     * @param signatureElement the element in the signature group of the pairing
+     * @param keyElement the element in the key group of the pairing
+     * @return a byte array representing the pairing
+     */
+    byte[] displayPairing(final GroupElement signatureElement, final GroupElement keyElement);
```
cody-littley, line 72: Why are we representing a string with a byte array?

Is the string intended to be something that's human readable? If yes, then I'd suggest using String objects in java if at all possible. If it isn't human readable, perhaps we should call it something other than a string.

hedera-bls-api/src/main/java/com/hedera/platform/bls/api/Field.java
```diff
@@ -0,0 +1,75 @@
...
+ * <p>This is a factory interface, responsible for creating {@link FieldElement field elements}
+ */
+public interface Field {
+    /**
+     * Creates a new field element from a long
+     *
+     * @param inputLong the long to use to create the field element
+     * @return the new field element
+     */
+    FieldElement elementFromLong(final long inputLong);
```
cody-littley: nit, `final` is not needed in interfaces for methods that don't have implementations. (Still use them for default methods though.)
```suggestion
    FieldElement elementFromLong(long inputLong);
```

hedera-bls-api/src/main/java/com/hedera/platform/bls/api/Field.java
```diff
@@ -0,0 +1,75 @@
...
+     */
+    FieldElement oneElement();
+
+    /**
+     * Creates a field element from a seed (256 bits)
+     *
+     * @param seed a seed to use to generate randomness
+     * @return the new field element
+     */
+    FieldElement randomElement(final byte[] seed);
```
cody-littley:
```suggestion
    FieldElement randomElement(byte[] seed);
```

hedera-bls-api/src/main/java/com/hedera/platform/bls/api/BilinearMap.java
```diff
@@ -0,0 +1,73 @@
...
+     * @param keyElement1 the key group element of the first pairing
+     * @param signatureElement2 the signature group element of the second pairing
+     * @param keyElement2 the key group element of the second pairing
+     * @return true if the 2 pairings have the same result, otherwise false
+     */
+    boolean comparePairing(
+            final GroupElement signatureElement1,
+            final GroupElement keyElement1,
+            final GroupElement signatureElement2,
+            final GroupElement keyElement2);
```
cody-littley:
```suggestion
    boolean comparePairing(
            GroupElement signatureElement1,
            GroupElement keyElement1,
            GroupElement signatureElement2,
            GroupElement keyElement2);
```

hedera-bls-api/src/main/java/com/hedera/platform/bls/api/BilinearMap.java
cody-littley, line 72:
```suggestion
    byte[] displayPairing(GroupElement signatureElement, GroupElement keyElement);
```

hedera-bls-api/src/main/java/com/hedera/platform/bls/api/FieldElement.java
```diff
@@ -0,0 +1,82 @@
...
+     */
+    byte[] toBytes();
+
+    /**
+     * Adds another field element to this one
+     *
+     * @param other the other field element
+     * @return a new field element which is the sum of this element and another
+     */
+    FieldElement add(final FieldElement other);
```
cody-littley: remove `final`s from this interface

hedera-bls-api/src/main/java/com/hedera/platform/bls/api/Group.java
```diff
@@ -0,0 +1,80 @@
...
+ */
+public interface Group {
+    /**
+     * Creates a new group element with value 1
+     *
+     * @return the new group element
+     */
+    GroupElement oneElement();
+
+    /** Creates a group element from a seed (256 bits) */
```
cody-littley: Why not just say 32 bytes?

hedera-bls-api/src/main/java/com/hedera/platform/bls/api/Group.java
```diff
@@ -0,0 +1,80 @@
...
+    GroupElement randomElement(final byte[] seed);
```
cody-littley: remove `final`s

hedera-bls-api/src/main/java/com/hedera/platform/bls/api/GroupElement.java
```diff
@@ -0,0 +1,84 @@
...
+     */
+    byte[] toBytes();
+
+    /**
+     * Takes the group element to the power of a field element
+     *
+     * @param exponent the field element exponent
+     * @return a new group element which is this group element to the power of a field element
+     */
+    GroupElement power(final FieldElement exponent);
```
cody-littley: remove `final`s

hedera-bls-impl/src/main/java/com/hedera/platform/bls/impl/BLS12381BilinearMap.java
```diff
@@ -0,0 +1,118 @@
...
+            throw new IllegalArgumentException("keyElement must be a BLS12381Group2Element");
+        }
+
+        // display output is always this size
+        final byte[] output = new byte[1249];
+
+        final int errorCode;
+        if ((errorCode = BLS12381Bindings.pairingDisplay(signature, key, output)) != SUCCESS) {
+            throw new BLS12381Exception("pairingDisplay", errorCode);
+        }
```
cody-littley: I see this pattern all over the place. Is there a reason why we couldn't instead do this? Doing the assignment in the if() statement doesn't save us any lines of code, but in my opinion makes the code harder to read.

```suggestion
        final int errorCode = BLS12381Bindings.pairingDisplay(signature, key, output);
        if (errorCode != SUCCESS) {
            throw new BLS12381Exception("pairingDisplay", errorCode);
        }
```

hedera-bls-impl/src/main/java/com/hedera/platform/bls/impl/BLS12381Group1Element.java
```diff
@@ -0,0 +1,196 @@
...
+
+    /** {@inheritDoc} */
+    @Override
+    public GroupElement compress() {
+        // Already compressed, no need to do anything
+        if (compressed) {
+            return this;
+        }
+
+        byte[] newGroupElement = new byte[GROUP.getCompressedSize()];
```
cody-littley:
```suggestion
        final byte[] newGroupElement = new byte[GROUP.getCompressedSize()];
```

hedera-bls-impl/src/main/java/com/hedera/platform/bls/impl/BLS12381Group2Element.java
```diff
@@ -0,0 +1,196 @@
...
+
+    /** {@inheritDoc} */
+    @Override
+    public GroupElement compress() {
+        // Already compressed, no need to do anything
+        if (compressed) {
+            return this;
+        }
+
+        byte[] newGroupElement = new byte[GROUP.getCompressedSize()];
```
cody-littley:
```suggestion
        final byte[] newGroupElement = new byte[GROUP.getCompressedSize()];
```

hedera-bls-impl/src/main/java/com/hedera/platform/bls/impl/LibraryLoader.java
```diff
@@ -0,0 +1,363 @@
...
+
+    /**
+     * Loads library from the current JAR archive and registers the native methods of the provided
+     * class. The library will be loaded at most once.
+     *
+     * <p>The file from JAR is copied into system temporary directory and then loaded. The temporary
+     * file is deleted after exiting.
+     */
+    public static void loadBundledLibrary(final Class<?> clazz) throws IOException {
+        Path pathInJar = getLibraryPathInResources();
```
cody-littley, line 66:
```suggestion
        final Path pathInJar = getLibraryPathInResources();
```

hedera-bls-impl/src/main/java/com/hedera/platform/bls/impl/LibraryLoader.java
```diff
@@ -0,0 +1,363 @@
...
+
+        try {
+            final File resourceFile = copyToTempDirectory(pathInJar, clazz);
+            System.load(resourceFile.getAbsolutePath());
+        } catch (final URISyntaxException e) {
+            logger.debug(e);
```
cody-littley, line 72: After this merges, we will need to add log markers to all log statements (else they won't show up in platform logs)

hedera-bls-impl/src/main/java/com/hedera/platform/bls/impl/LibraryLoader.java
```diff
@@ -0,0 +1,363 @@
...
+
+    /**
+     * Copies a file into a temporary directory regardless of if it is in a JAR or not.
+     *
+     * @param relativePath A relative path to a file or directory, relative to the resources folder.
+     * @return The file or directory you want to load.
+     * @throws IOException If at any point processing of the resource file fails.
+     * @throws URISyntaxException If resource file cannot be found
+     */
+    private static File copyToTempDirectory(Path relativePath, Class<?> outsideClass)
```
cody-littley, line 84:
```suggestion
    private static File copyToTempDirectory(final Path relativePath, final Class<?> outsideClass)
```

hedera-bls-impl/src/main/java/com/hedera/platform/bls/impl/LibraryLoader.java
```diff
@@ -0,0 +1,363 @@
...
+        tempDirectory.mkdirs();
+
+        // Is the user loading resources that are from inside a JAR?
+        Path fullJarPathURL =
+                Path.of(ResourceLoader.getThePathToTheJarWeAreIn(outsideClass).toURI());
+
+        // Test if we are in a JAR
+        if (isJarFile(fullJarPathURL)) {
+            File extracted =
+                    nestedExtract(tempDirectory, fullJarPathURL.resolve(relativePath).toString());
```
cody-littley, line 99:
```suggestion
        final File tempDirectory = ResourceLoader.createMainTempDirectory();

        // Create the required directories.
        tempDirectory.mkdirs();

        // Is the user loading resources that are from inside a JAR?
        final Path fullJarPathURL =
                Path.of(ResourceLoader.getThePathToTheJarWeAreIn(outsideClass).toURI());

        // Test if we are in a JAR
        if (isJarFile(fullJarPathURL)) {
            final File extracted =
                    nestedExtract(tempDirectory, fullJarPathURL.resolve(relativePath).toString());
```

hedera-bls-impl/src/main/java/com/hedera/platform/bls/impl/LibraryLoader.java
cody-littley, line 84: Java has this strange duality between `Path` and `File`. For the most part, think of `File` as the old fashioned API that we can't get rid of due to momentum, and `Path` as the newer and better API.

If at all possible, I suggest using the `Path` APIs. For the most part, the entirety of the platform code base has been migrated to use the `Path` API.

hedera-bls-impl/src/main/java/com/hedera/platform/bls/impl/LibraryLoader.java
```diff
@@ -0,0 +1,363 @@
...
+        } else {
+            file = new File(url.getFile());
+        }
+
+        if (!file.isFile()) {
+            throw new IOException("resource to load should be a file");
+        }
+
+        File resource = new File(relativePath.toUri());
+        File resourceCopiedToTempFolder = new File(outputDirectory, resource.getName());
```
cody-littley, line 164:
```suggestion
        final File resource = new File(relativePath.toUri());
        final File resourceCopiedToTempFolder = new File(outputDirectory, resource.getName());
```

hedera-bls-impl/src/main/java/com/hedera/platform/bls/impl/LibraryLoader.java
```diff
@@ -0,0 +1,363 @@
...
+    /**
+     * If the string does not start with a slash, then make sure it does.
+     *
+     * @param s A string to prefix
+     * @return A string with a slash prefixed
+     */
+    private static String prefixStringWithSlashIfNotAlready(String s) {
+        if (!s.startsWith("/")) {
+            s = "/" + s;
+        }
```
cody-littley, line 179:
```suggestion
    private static String prefixStringWithSlashIfNotAlready(final String s) {
        if (!s.startsWith("/")) {
            s = "/" + s;
        }
```

This is the kind of stuff that is handled more elegantly by the path APIs.

hedera-bls-impl/src/main/java/com/hedera/platform/bls/impl/LibraryLoader.java
```diff
@@ -0,0 +1,363 @@
...
+     * @return The final extracted file.
+     * @throws IOException
+     * @throws URISyntaxException
+     */
+    private static File nestedExtract(File extractTo, String fullPath) throws IOException {
+        final String JAR = ".jar";
+
+        // After this line we have something like
+        // file:C/app, some/lazysodium, file.txt
+        String[] split = fullPath.split("(\\.jar/)");
```
cody-littley, line 198:
```suggestion
    private static File nestedExtract(final File extractTo, final String fullPath) throws IOException {
        final String JAR = ".jar";

        // After this line we have something like
        // file:C/app, some/lazysodium, file.txt
        final String[] split = fullPath.split("(\\.jar/)");
```

hedera-bls-impl/src/main/java/com/hedera/platform/bls/impl/LibraryLoader.java
```diff
@@ -0,0 +1,363 @@
...
+        }
+
+        // We have no ".jar/" so we go straight
+        // to extraction.
+        if (split.length == 1) {
+            logger.debug("Extracted {} to {}", fullPath, extractTo.getAbsolutePath());
+            return extractFilesOrFoldersFromJar(extractTo, new URL(fullPath), "");
+        }
+
+        String currentExtractionPath = "";
```
cody-littley, line 212: I'll stop commenting on each one, but there are bunch of variables that need to be final in this class.

hedera-bls-impl/src/main/java/com/hedera/platform/bls/impl/LibraryLoader.java
```diff
@@ -0,0 +1,363 @@
...
+        File extracted = null;
+        File nestedExtractTo = extractTo;
+        for (int i = 0; i < split.length - 1; i++) {
+            // Remember part = "file:C/app". But we need to know
+            // where to extract these files. So we have
+            // to prefix it with the current extraction path. We can't
+            // just dump everything in the temp directory all the time.
+            // Of course, we also suffix it with a ".jar". So at the end,
+            // we get something like "file:C:/temp/app.jar"
+            String part = currentExtractionPath + split[i] + JAR;
```
cody-littley, line 222: How has this windows specific code been tested? Also, I'm pretty sure `Path` handles this kind of parsing for us.

hedera-bls-impl/src/main/java/com/hedera/platform/bls/impl/Utils.java
```diff
@@ -0,0 +1,40 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.hedera.platform.bls.impl;
+
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
+
+public final class Utils {
```
cody-littley, line 22: short javadoc

cody-littley (approved):

buildSrc/src/main/kotlin/com.hedera.platform.rust-library.gradle.kts
litt3: seems to be a gradle necessity - deleting breaks the build

hedera-bls-api/src/main/java/com/hedera/platform/bls/api/BLSLoader.java
litt3: Sounds good. Made an issue https://github.com/hashgraph/hedera-bls-cryptography/issues/8

litt3: Something to consider - the returned object is *not* stateful in any way

hedera-bls-api/src/main/java/com/hedera/platform/bls/api/BilinearMap.java
litt3: String was the wrong term - changed to "byte array". Nothing human readable about this

hedera-bls-api/src/main/java/com/hedera/platform/bls/api/Field.java
litt3: fixed throughout

hedera-bls-api/src/main/java/com/hedera/platform/bls/api/Field.java
litt3: fixed

hedera-bls-api/src/main/java/com/hedera/platform/bls/api/BilinearMap.java
litt3: fixed

hedera-bls-api/src/main/java/com/hedera/platform/bls/api/BilinearMap.java
litt3: fixed

hedera-bls-api/src/main/java/com/hedera/platform/bls/api/FieldElement.java
litt3: fixed

hedera-bls-api/src/main/java/com/hedera/platform/bls/api/Group.java
litt3: Bits on the brain. fixed

hedera-bls-api/src/main/java/com/hedera/platform/bls/api/Group.java
litt3: done

hedera-bls-api/src/main/java/com/hedera/platform/bls/api/GroupElement.java
litt3: done

hedera-bls-impl/src/main/java/com/hedera/platform/bls/impl/BLS12381BilinearMap.java
litt3: It's a smell from a previous layout. Fixed throughout

hedera-bls-impl/src/main/java/com/hedera/platform/bls/impl/BLS12381Group1Element.java
litt3: fixed

hedera-bls-impl/src/main/java/com/hedera/platform/bls/impl/BLS12381Group2Element.java
litt3: fixed

hedera-bls-impl/src/main/java/com/hedera/platform/bls/impl/LibraryLoader.java
litt3: As we discussed, I'm holding off on polishing this class until we're more certain that this method of library loading will persist

hedera-bls-impl/src/main/java/com/hedera/platform/bls/impl/Utils.java
litt3: added

poulok (approved):

nathanklick (approved):

PR #19, opened 2024-07-17
ci: add initial ci pipeline and gradle implementations

### LINKED ISSUE #12: Add initial CI pipeline and gradle implementations
## Description

Add an initial CI pipeline implementation to this repository. The implementation should at a minimum build the Rust native code for both Linux x86_64 and Aarch64 architectures.

rbarker-dev:
**Description**:

Adds initial CI/CD pipelines for hedera-cryptography repository

**Related issue(s)**:

Fixes #12


nathanklick (changes requested):
Maven Group ID should be `com.hedera.cryptography` and not `com.hedera` and the package/module prefix should be `com.hedera.cryptography`

jjohannes (approved):
@rbarkerSL and I worked on most of this together. :+1: from me.

Note for future reference: We copied `gradle/plugins` with only little adjustments from https://github.com/hashgraph/hedera-services. There might be configuration in there not needed in this project. This will be consolidated as part of https://github.com/hashgraph/hedera-services/issues/14255.

.github/CODEOWNERS
```diff
@@ -0,0 +1,44 @@
+###################################
```
mxtartaglia-sl, line 1: Can we include platform-team as the owner of each of the modules?

build.gradle.kts
```diff
@@ -13,6 +13,8 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+
+/*
```
mxtartaglia-sl: Why would we want to keep this commented code? maybe remove it if not necessary?

hedera-cryptography-altbn128/build.gradle.kts
```diff
@@ -0,0 +1,25 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+plugins {
+    id("com.hedera.gradle.services")
```
mxtartaglia-sl: What are the implications of publishing with "com.hedera.gradle.services"?
As this is all to be maintained by the platform team, and to be used by the platform code

hedera-nativesupport/build.gradle.kts
```diff
@@ -0,0 +1,25 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+plugins {
+    id("com.hedera.gradle.services")
```
mxtartaglia-sl: In this case it seems that the group will be `com.hedera`
@nathanklick suggested `com.hedera.commons` for this module

gradle/plugins/src/main/kotlin/com.hedera.gradle.services-publish.gradle.kts
```diff
@@ -0,0 +1,53 @@
...
+                    name = "Hedera Services Team"
+                    email = "hedera-services@swirldslabs.com"
+                    organization = "Hedera Hashgraph"
+                    organizationUrl = "https://www.hedera.com"
+                }
+                developer {
+                    name = "Hedera Smart Contracts Team"
+                    email = "hedera-smart-contracts@swirldslabs.com"
+                    organization = "Hedera Hashgraph"
+                    organizationUrl = "https://www.hedera.com"
```
mxtartaglia-sl: I see this is the plugin being used by pairings, pairings impl, native support, etc:

What about changing the name of this plugin to avoid confusion with the one in hedera-services and the distinction between `services` and `platform` code there?

Also, I don't see the smart contacts or services teams interacting with this repo much, but the Platform team needs to be in. Perhaps the code is more similar to the `gradle/plugins/src/main/kotlin/com.hedera.gradle.platform-publish.gradle.kts` file?

gradle/plugins/src/main/kotlin/com.hedera.gradle.services.gradle.kts
```diff
@@ -0,0 +1,22 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+plugins {
+    id("java-library")
+    id("com.hedera.gradle.java")
+}
+
+group = "com.hedera.cryptography"
```
mxtartaglia-sl: Still I think that we would need a different one for publishing 'native-support'
as @nathanklick suggested

mxtartaglia-sl (commented):
Adding some comments and questions.

.github/CODEOWNERS
rbarker-dev: Sure thing; I'll get that added in

gradle/plugins/src/main/kotlin/com.hedera.gradle.services.gradle.kts
rbarker-dev: moving to native-support breaks things based on the `native` keyword in java. I recommend leaving it as `nativesupport`

gradle/plugins/src/main/kotlin/com.hedera.gradle.services-publish.gradle.kts
rbarker-dev: I moved the module dependencies to use com.hedera.gradle.platform-publish.gradle.kts

hedera-cryptography-altbn128/build.gradle.kts
rbarker-dev: Swapped these over to platform.

gradle/plugins/src/main/kotlin/com.hedera.gradle.services.gradle.kts
mxtartaglia-sl: Sorry @rbarkerSL, I was referring to my previous comment.
I meant that for the lib using this plugin (wrote `native-support` but should have been `hedera-nativesupport`) , @nathanklick  suggested usign the group:`com.hedera.commons`

.github/CODEOWNERS
```diff
@@ -0,0 +1,52 @@
...
+# NOTE: This rule is overriden by the more specific rules below. This is the catch-all rule for all files not covered by the more specific rules below.
+*                                               @hashgraph/devops-ci @hashgraph/release-engineering-managers
+
+#########################
+####   Java Modules  ####
+#########################
+/hedera-cryptography-altbn128/                  @hashgraph/platform-hashgraph @hashgraph/platform-base
+/hedera-cryptography-pairings-api/              @hashgraph/platform-hashgraph @hashgraph/platform-base
+/hedera-cryptography-pairings-signatures/       @hashgraph/platform-hashgraph @hashgraph/platform-base
+/hedera-common-nativesupport/                          @hashgraph/platform-hashgraph @hashgraph/platform-base
```
nathanklick: nit: Spacing appears to differ here, but could be a GH UI issue.

.github/dependabot.yml
```diff
@@ -0,0 +1,18 @@
...
+  - package-ecosystem: gradle
+    directory: "/"
+    schedule:
+      interval: weekly
+    open-pull-requests-limit: 10
+  - package-ecosystem: gradle
+    directory: "/gradle/plugins"
+    schedule:
+      interval: weekly
+    open-pull-requests-limit: 10
```
nathanklick: I don't believe this is required since I believe dependabot scans for all gradle projects under the specified path of `/` in the prior configuration.

.github/workflows/build.yml
```diff
@@ -0,0 +1,73 @@
+name: Build
+
+on:
+  push:
+    branches:
+      - main
+      - develop
```
nathanklick:
```suggestion
```

.github/workflows/build.yml
```diff
@@ -0,0 +1,73 @@
...
+      - release/*
+  pull_request:
+    branches:
+      - main
+      - develop
```
nathanklick:
```suggestion
```

mxtartaglia-sl (approved):

.github/CODEOWNERS
rbarker-dev: It's not; added `-common` to the module but forgot to remove the equivalent whitespace.

mxtartaglia-sl:
Are we using main for this repo instead of develop?

rbarker-dev:
> Are we using main for this repo instead of develop?

Yes.

mxtartaglia-sl (approved):

mxtartaglia-sl:
@rbarkerSL when I add the module-info in hedera-common-nativesupport
```
/**
 * Provides a set of generic functions loading native libraries in different system architectures when packaged
 * in a jar using a predefined organization, so they can be accessed with JNI.
 */
module com.hedera.common.nativesupport {
    exports com.hedera.common.nativesupport;

    requires static transitive com.github.spotbugs.annotations;
}

```
I receive the following error:
```
Could not determine the dependencies of task ':hedera-common-nativesupport:compileJava'.
> Could not resolve all dependencies for configuration ':hedera-common-nativesupport:compileClasspath'.
   > Could not find com.github.spotbugs:spotbugs-annotations:.
     Required by:
         project :hedera-common-nativesupport

Possible solution:
 - Declare repository providing the artifact, see the documentation at https://docs.gradle.org/current/userguide/declaring_repositories.html
```

https://github.com/hashgraph/hedera-cryptography#:~:text=14136_native_support

jjohannes:
@mxtartaglia-sl The problem with `Could not find com.github.spotbugs:spotbugs-annotations:.` is that you do not have a version defined for it. The generic Gradle error message is super unhelpful if you don't know that. the idication is the `:.` in the end (no version after the `:`).

You need to add the version here. Same as we do it in `hedera-services`:
[hedera-dependency-versions/build.gradle.kts](https://github.com/hashgraph/hedera-cryptography/blob/12-add-initial-ci-pipeline-and-gradle-implementations/hedera-dependency-versions/build.gradle.kts)

As example, I added the version for `spotbugs-annotations` in this PR.

```
    api("com.github.spotbugs:spotbugs-annotations:4.7.3") {
        because("com.github.spotbugs.annotations")
    }
```

jjohannes:
@mxtartaglia-sl @rbarkerSL @nathanklick after reading through the reviews, I pushed an update to this PR:

- I removed the part from the Gradle configs that have nothing to do with this repo. I realized that it's super confusing to those who miss context if we have them here just to remove them in the next step.
- Now that we actually have multiple "groups", it already starts to get confusing when setting them sometimes in plugins, sometimes in `build.gradle.kts`. Therefore, I changed the setup so that all the `group`s for the Modules are now defined in `settings.gradle.kts`. This moves them out of the plugins and makes the plugins more general reusable across repositories/products (already having in mind https://github.com/hashgraph/hedera-services/issues/14255). I want to do the same in _hedera-services_ - for more context, how and why we can do this now: https://github.com/hashgraph/hedera-services/pull/14245

The whole definition now looks like this

```
javaModules {
    directory("common") {
        group ="com.hedera.common"
    }
    directory("cryptography") {
        group = "com.hedera.cryptography"
    }
    versions("hedera-dependency-versions")
}
```

In this setup the `module-info.java` files are discovered and you do not need to list all the modules individually. For that, I moved the modules into a subfolder according to their "group". It feels natural to have all modules that belong to one "group" also grouped in one subfolder. Maybe a pattern we should eventually use in all repos (_hedera-services_ kind of has this structure already).

```
├── common
│   └── hedera-common-nativesupport
├── cryptography
│   ├── hedera-cryptography-altbn128
│   ├── hedera-cryptography-pairings-api
│   └── hedera-cryptography-pairings-signatures

```

Let me know if you agree. If not, I can revert/change that.

jjohannes (approved):

mxtartaglia-sl (approved):
Thanks @jjohannes and @rbarkerSL!
Sorry, I did not check if the versions were available; I assumed the version file was the same as in `hedera-services.`

The changes look good. This is really good, and I see that there is a path to a final version without repeated steps (like defining a library version in all repositories) or different structures per repository in which one has to remember specific rules; we can keep building on top of this base stone.
Thanks for completing this.

.github/workflows/build.yml
```diff
@@ -0,0 +1,71 @@
...
+      - name: Gradle Unit Test
+        run: ./gradlew :reports:testCodeCoverageReport --scan --continue
+
+      - name: Publish Maven Central
+        env:
+          NEXUS_USERNAME: ${{ secrets.SVCS_OSSRH_USERNAME }}
+          NEXUS_PASSWORD: ${{ secrets.SVCS_OSSRH_PASSWORD }}
+          SIGNING_KEY: ${{ secrets.GPG_KEY_CONTENTS }}
+          SIGNING_PASSPHRASE: ${{ secrets.GPG_KEY_PASSPHRASE }}
+        run: ./gradlew releaseMavenCentralSnapshot -PpublishingPackageGroup=com.hedera --scan -PpublishSigningEnabled=true --no-configuration-cache
```
nathanklick: We should not be publishing to Maven Central on every pull request and commit against those PRs. This will get us rate limited very quickly and will impact our ability to push critical code from our self-hosted runners.

Publishing to Maven Central should be done sparingly (eg: only on merge to the `main` branch).

.github/workflows/build.yml
```diff
@@ -0,0 +1,71 @@
...
+        run: ./gradlew assemble --scan
+
+      - name: Spotless Check
+        run: ./gradlew spotlessCheck --scan --continue
+
+      - name: Dependency Scopes Check
+        run: ./gradlew checkAllModuleInfo --scan --continue
+
+      - name: Gradle Unit Test
+        run: ./gradlew :reports:testCodeCoverageReport --scan --continue
```
nathanklick: This does not guarantee the unit tests will be executed. This only runs the JaCoCo code coverage aggregation.

nathanklick (changes requested):

.github/workflows/build.yml
nathanklick, line 1: This workflow should also use a unit test reporting plugin like `step-security/publish-unit-test-result-action`.

nathanklick (changes requested):

gradle/plugins/src/main/kotlin/com/hedera/gradle/spotless/StripOldLicenseFormatterStep.kt
nathanklick, line 1: Probably not needed here since this is all new code.

.github/workflows/build.yml
jjohannes: In the setup we have now it does. There is a hard dependency from the report generation tasks to the test tasks via the dependency from the report aggregation project to all test-report-producing projects. 
This changed when we introduced the single report aggregation project/task.

But for clarity, it would not hurt to add "test" as additional argument.

jjohannes:
@mxtartaglia-sl Centralizing third party versions across multiple repos is another topic to discuss in the future. After we unified Gradle setup unification (https://github.com/hashgraph/hedera-services/issues/14255) I would say.

There is the option to publish what we have in `heder-services/hedera-dependency-versions` as a BOM (metadata file with versions). Similar to: https://repo1.maven.org/maven2/org/springframework/boot/spring-boot-dependencies/3.3.2/
And then use that across multiple repos.

But in this area, I feel like accepting duplication and relying on Dependabot to keep things on the latest versions everywhere is also an option.

I think the experience you gain when working with the setup as it is now will be helpful to discuss options in a few weeks.

gradle/plugins/src/main/kotlin/com/hedera/gradle/spotless/StripOldLicenseFormatterStep.kt
rbarker-dev: Removed

.github/workflows/build.yml
rbarker-dev: Added, probably needs some additional configuration

.github/workflows/build.yml
rbarker-dev: Added the test argument

.github/workflows/build.yml
rbarker-dev: Added a check to ensure that the ref is `main`

.github/workflows/build.yml
```diff
@@ -0,0 +1,83 @@
...
+
+      # only publish on merge to main
+      - name: Publish Maven Central
+        if: ${{ github.ref == 'refs/heads/main' }}
+        env:
+          NEXUS_USERNAME: ${{ secrets.SVCS_OSSRH_USERNAME }}
+          NEXUS_PASSWORD: ${{ secrets.SVCS_OSSRH_PASSWORD }}
+          SIGNING_KEY: ${{ secrets.GPG_KEY_CONTENTS }}
+          SIGNING_PASSPHRASE: ${{ secrets.GPG_KEY_PASSPHRASE }}
+        run: ./gradlew releaseMavenCentralSnapshot -PpublishingPackageGroup=com.hedera --scan -PpublishSigningEnabled=true --no-configuration-cache
```
nathanklick:
```suggestion
        run: ./gradlew releaseMavenCentralSnapshot -PpublishingPackageGroup=com.hedera.cryptography --scan -PpublishSigningEnabled=true --no-configuration-cache
```

nathanklick (changes requested):

gradle/plugins/src/main/kotlin/com.hedera.gradle.hedera-cryptography-publish.gradle.kts
```diff
@@ -0,0 +1,66 @@
...
+
+plugins {
+    id("java")
+    id("com.hedera.gradle.maven-publish")
+}
+
+// Publishing tasks are only enabled if we publish to the matching group.
+// Otherwise, Nexus configuration and credentials do not fit.
+val publishingPackageGroup = providers.gradleProperty("publishingPackageGroup").getOrElse("")
+
```
jjohannes, line 25: I don't think this split up is necessary. The publishing credentials are bound to "com.hedera" IIUC and not the "subgroups". In hedera-services we also have this and publish on on go: 'con.hedera.hashgraph' and 'con.hedera.evm'.

rbarker-dev: The request was to distinguish the two in order to future proof a little bit. There is a path where we pull nativesupport into its own repository. This will simplify doing that.

nathanklick (approved):

PR #36, opened 2024-08-06
feat: 13359 ECKeyGen

### LINKED ISSUE #13359: Create tool to render ECDH keys
Similar to `ssh-keygen` generate a tool that allows the creation of EC Private / Public key pairs.

The EC private key for each node must be stored in a `.pem` file, similar to how the RSA private key is stored in a `.pem` file.   The public portion of the EC key should be stored in a separate `.pem` file when generated by the tool.  

-----

### This task depends on:
- [ ] https://github.com/hashgraph/hedera-services/issues/13324
- [ ] https://github.com/hashgraph/hedera-services/issues/14100
- [ ] https://github.com/hashgraph/hedera-services/issues/13630

mxtartaglia-sl:
**Description:**
A tool similar to ssh-keygen that allows the creation of EC Private / Public key pairs.
**Note for reviewers(s):**
Compiled artifacts are provided temporally. They will be removed after CI/CD support for compiling rust code is available.

**Related issue(s):**

Fixes https://github.com/hashgraph/hedera-services/issues/13359

**Proposal this work is based on:**[ platform-sdk/docs/proposals/TSS-Library/TSS-Library.md](https://github.com/hashgraph/hedera-services/blob/develop/platform-sdk/docs/proposals/TSS-Library/TSS-Library.md)


cryptography/hedera-cryptography-eckeygen/src/main/java/com/hedera/cryptography/eckeygen/KeyGen.java
```diff
@@ -0,0 +1,74 @@
...
+ * Key generation tool
+ */
+public class KeyGen {
+
+    private static final KeysGenerationService KEYS_SERVICE = new KeysGenerationService(
+            SignatureSchema.create(Curve.ALT_BN128, GroupAssignment.GROUP1_FOR_SIGNING),
+            new NativeKeyGenerator().initialize());
+
+    public static void main(String[] args) throws Exception {
+        if (args.length == 0 || args[0].equals("--help")) {
```
kfa-aguda: I suggest collecting the args into well named "command name", "private key path", "public key path" etc. for readability.

cryptography/hedera-cryptography-eckeygen/src/main/java/com/hedera/cryptography/eckeygen/PemFiles.java
```diff
@@ -0,0 +1,105 @@
...
+/**
+ * Writes and reads the base64 string in Pem files according to <a href="https://datatracker.ietf.org/doc/html/rfc7468">...</a>
+ */
+public class PemFiles {
+
+    /**
+     * Reads the content of a PEM file.
+     *
+     * @param path The location of the file in the fileSystem
+     * @param keyType eiter "PUBLIC KEY" or "PRIVATE KEY" string
```
kfa-aguda:
```suggestion
     * @param keyType either "PUBLIC KEY" or "PRIVATE KEY" string
```

cryptography/hedera-cryptography-eckeygen/src/main/java/com/hedera/cryptography/eckeygen/PemFiles.java
```diff
@@ -0,0 +1,105 @@
...
+     * @return the base64 string contained in the PemFile
+     * @throws IOException In case of file reading error
+     */
+    @NonNull
+    public static String pemRead(@NonNull final String path, @NonNull final String keyType) throws IOException {
+        Objects.requireNonNull(keyType, "keyType must not be null");
+        String pemContent = Files.readString(Path.of(Objects.requireNonNull(path, "path must not be null")));
+
+        // Define PEM header and footer
+        String header = "-----BEGIN " + keyType + "-----";
```
kfa-aguda: is it worth validating the keyType is of the exact values required as per javadoc? I see you're passing the correct values from main but I think it's worth validating here too.

cryptography/hedera-cryptography-eckeygen/src/main/java/com/hedera/cryptography/eckeygen/PemFiles.java
```diff
@@ -0,0 +1,105 @@
...
+     * @param path The location of the file in the fileSystem
+     * @param keyType eiter "PUBLIC KEY" or "PRIVATE KEY" string
+     * @return the path of where the file was written
+     * @throws IOException In case of file reading error
+     */
+    @NonNull
+    public static Path pemWrite(
+            @NonNull final String path, @NonNull final String base64Key, @NonNull final String keyType)
+            throws IOException {
+        String header = "-----BEGIN " + Objects.requireNonNull(keyType, "keyType must not be null") + "-----\n";
```
kfa-aguda: same comment above applies.

cryptography/hedera-cryptography-eckeygen/src/main/java/com/hedera/cryptography/eckeygen/PemFiles.java
mxtartaglia-sl: Yes, agreed, will add an enum so there is no way to introduce an invalid value in the parameter.

cryptography/hedera-cryptography-eckeygen/src/main/java/com/hedera/cryptography/eckeygen/KeyGen.java
mxtartaglia-sl: Makes sense

cryptography/hedera-cryptography-eckeygen/src/main/java/com/hedera/cryptography/eckeygen/KeyGenerator.java
```diff
@@ -0,0 +1,43 @@
...
+public interface KeyGenerator {
+
+    /**
+     * Generate a key pair (private key and public key) and return them as strings.
+     *
+     * @param groupAssignment  An int representing the {@link GroupAssignment} ordinal for selecting the elliptic curve group to use.
+     * @param out  An array of size 2 to store the resulting private key and public key.
+     * @return an integer status code (0 for success, -1 for failure).
+     */
+    int generateKeyPair(final int groupAssignment, byte[][] out);
```
edward-swirldslabs: The `out` parameter is very confusing to me.  My best guess is that the input 2D array is only partially initialized, and it is expected to be a length of 2?  But the second subscript on each is uninitialized and or set to null and the method will set those values for a return?  

I'm curious why the return value couldn't be byte[][], and have it set to null if there is an error.     It's a lot cleaner and clearer if there is 1 source of responsibility for instantiating the 2D array.  The alternative is to create a class to hold the pair.

cryptography/hedera-cryptography-eckeygen/src/main/java/com/hedera/cryptography/eckeygen/KeyGenerator.java
```diff
@@ -0,0 +1,43 @@
...
+
+    /**
+     * Generate a public key given an existent private key and return it as string.
+     *
+     * @param groupAssignment  An int representing the {@link GroupAssignment} ordinal for selecting the elliptic curve group to use.
+     * @param sk  An array representing the private key.
+     * @return An array representing the public key
```
edward-swirldslabs: you have an integer status returned above.  Is it possible for this operation to fail as well?  is null possible as a return value to indicate a failure?   since `byte[]` can be null, we need to annotated the return value with a nullity marking.

cryptography/hedera-cryptography-eckeygen/src/main/java/com/hedera/cryptography/eckeygen/NativeKeyGenerator.java
```diff
@@ -0,0 +1,73 @@
...
+    private static final AtomicBoolean IS_INITIALIZED = new AtomicBoolean(false);
+
+    /**
+     * Initializes the class by loading the necessary native libraries.
+     *
+     * @return this instance.
+     */
+    @NonNull
+    public NativeKeyGenerator initialize() {
+        if (IS_INITIALIZED.compareAndSet(false, true)) {
```
edward-swirldslabs: Thread A and Thread B hit line 40 at the same time.  Thread A sets the new value successfully and enters the body of the if block and does the work of initializing.    Thread B wakes up, sees the value is already true, skips the block and immediately returns.   Thread B returns before Thread A is done with initialization.  This looks like a race condition. 

I think the initialization has to be done while thread A is holding onto the lock.

cryptography/hedera-cryptography-eckeygen/src/test/java/com/hedera/cryptography/eckeygen/KeysGenerationServiceTest.java
```diff
@@ -0,0 +1,88 @@
...
+
+import com.hedera.cryptography.eckeygen.KeysGenerationService.KeysServiceException;
+import com.hedera.cryptography.pairings.api.Curve;
+import com.hedera.cryptography.pairings.signatures.api.GroupAssignment;
+import com.hedera.cryptography.pairings.signatures.api.SignatureSchema;
+import java.util.Base64;
+import org.junit.jupiter.api.Test;
+import org.mockito.ArgumentMatchers;
+
+class KeysGenerationServiceTest {
```
edward-swirldslabs, line 34: Is it possible to test any of the key functionality such as signing and encrypting?  Or not yet?

mxtartaglia-sl: Not yet; the scope here was to generate the key files. The operations for signing and encrypting will be generated in future tasks and in the 'pairings-signatures-library'.

cryptography/hedera-cryptography-eckeygen/src/main/java/com/hedera/cryptography/eckeygen/KeyGenerator.java
mxtartaglia-sl: > since byte[] can be null, we need to annotated the return value with a nullity marking.

I can't annotate. This would produce many complications when compiling the header file, as it would require all the necessary libraries to be in the classpath.

But i can do a better job in the javadocs

cryptography/hedera-cryptography-eckeygen/src/main/java/com/hedera/cryptography/eckeygen/KeyGenerator.java
mxtartaglia-sl: > My best guess is that the input 2D array is only partially initialized, and it is expected to be a length of 2? But the second subscript on each is uninitialized and/or set to null, and will the me set those values for a return?

 Correct!

> It's a lot cleaner and clearer if there is 1 source of responsibility for instantiating the 2D array

I agree. I wouldn't say I like this, but this approach facilitates the JNI operations. Returning Java types in JNI methods creates a coupling between Rust and Java in the form of strings. We need to qualify type names fully, and in case something changes, we have to maintain that parity. I want to reduce that coupling. Furthermore, returning a Java matrix in Rust requires too much boilerplate Rust code, which I am saving by sending the outmost array as an input parameter.

I'll review this to see if i can circumvent this, or restate the javadocs to be clearer

cryptography/hedera-cryptography-eckeygen/src/main/java/com/hedera/cryptography/eckeygen/NativeKeyGenerator.java
mxtartaglia-sl: Although no multithreading is expected here, I will fix this issue as it can become a template for further solutions.
I don't want to pay the cost for a synchronized method in all calls that will mostly be:
'if (false){} return;'

cryptography/hedera-cryptography-eckeygen/Cargo.toml
```diff
@@ -0,0 +1,31 @@
...
+
+[dependencies]
+jni = "0.21.1"
+ark-std = { version = "0.4.0" }
+ark-ec = { version = "0.4.0" }
+ark-ff = { version = "0.4.0" }
+ark-bn254 = { version = "0.4.0", default-features = false, features = [ "curve" ] }
+ark-serialize = { version = "^0.4.0", default-features = true }
+
+rand = "*"
```
rbarker-dev: Let's throw a newline at the end here

cryptography/hedera-cryptography-eckeygen/src/main/resources/software/darwin/arm64/libkey_gen.dylib
rbarker-dev, line 1: Why are we committing binary files?

cryptography/hedera-cryptography-eckeygen/src/main/resources/software/linux/amd64/libkey_gen.so
rbarker-dev, line 1: Why are we committing binary files?

rbarker-dev (changes requested):

cryptography/hedera-cryptography-eckeygen/src/main/resources/software/darwin/arm64/libkey_gen.dylib
rbarker-dev: until we have support for compiling Rust, need to push them so that the tests can execute and the code works.

rbarker-dev (dismissed):

cryptography/hedera-cryptography-eckeygen/src/main/java/com/hedera/cryptography/eckeygen/KeyGenerator.java
mxtartaglia-sl: Ok, I returned the byte[][] from rust code. 
It required more bureaucratic code from the Rust side, plus handling of lifespans for the returning variable, but it ended up with a more consistent approach between the two methods.
Indeed, the java side got clearer and cleaner.

cryptography/hedera-cryptography-eckeygen/src/main/java/com/hedera/cryptography/eckeygen/KeyGenerator.java
mxtartaglia-sl: I introduced the annotations; the effect is that it is more difficult to compile the headers from the current version. But I should not require them to be compiled so often. Besides, it is an optional step; what matters is the implementation.
In any case, if needed, I can create a new class using only the native methods to generate the new headers and then throw out the new class.

cryptography/hedera-cryptography-eckeygen/src/main/java/com/hedera/cryptography/eckeygen/NativeKeyGenerator.java
mxtartaglia-sl: Let me know what you think about the new approach.

edward-swirldslabs (approved):
LGTM

rbarker-dev (approved):

PR #45, opened 2024-08-24
feat: 14788 group2 api impl ops

### LINKED ISSUE #14788: Implement Group2 Group Elements pairings-api-impl operations in JNI and rust
Following: https://github.com/hashgraph/hedera-services/blob/develop/platform-sdk/docs/proposals/TSS-Library/TSS-Library.md 

Implement Group 2 and Group2Elements operations for Alt bn 128 curve.
Thorough unit testing and performance testing are deferred for a latter point

mxtartaglia-sl:
**Description**:
Group 2 and Group2Elements operations.
Additional change of how FieldElements are represented in BigInteger

**Related issue(s)**:

Fixes [#14788](https://github.com/hashgraph/hedera-services/issues/14788)

**Notes for reviewer**:
Binary files are provided until compiling support with gradle for rust is completed.

**Checklist**


cryptography/hedera-cryptography-altbn128/src/main/java/com/hedera/cryptography/altbn128/adapter/Group2LibraryAdapter.java
```diff
@@ -31,8 +31,8 @@ public interface Group2LibraryAdapter {
     int SUCCESS = 0;
     /** The return code that represents that the requested point is not in the curve */
     int NOT_IN_CURVE = -4;
-
-    int POINT_BYTE_SIZE = 32;
+    /** The coordinate size */
+    int POINT_COORDINATE_SIZE = 32;
```
lpetrovic05: 32 bytes? bits? add a comment stating it

cryptography/hedera-cryptography-altbn128/src/main/java/com/hedera/cryptography/altbn128/AltBn128Group2.java
```diff
@@ -0,0 +1,177 @@
...
+import java.util.Collection;
+import java.util.List;
+import java.util.Objects;
+
+/**
+ * The implementation of a {@link Group2}
+ * for {@link com.hedera.cryptography.pairings.api.curves.KnownCurves#ALT_BN128}
+ */
+public class AltBn128Group2 implements Group {
+    private final Group2 facade;
```
lpetrovic05: a struggling to figure out what `Group2` is. The javadoc says this is an implementation of `Group2`, but the interface it implements is `Group`

cryptography/hedera-cryptography-altbn128/src/main/java/com/hedera/cryptography/altbn128/AltBn128Group2.java
```diff
@@ -0,0 +1,177 @@
...
+    @NonNull
+    @Override
+    public GroupElement batchAdd(@NonNull final Collection<GroupElement> elements) {
+        Objects.requireNonNull(elements, "elements must not be null");
+        if (elements.stream().anyMatch(e -> !AltBn128Group2Element.class.isAssignableFrom(e.getClass()))) {
+            throw new IllegalArgumentException("elements must implement AltBn128Group2Element");
+        }
+        List<AltBn128Group2Element> elems =
+                elements.stream().map(AltBn128Group2Element.class::cast).toList();
+        byte[][] all = new byte[elems.size()][];
```
lpetrovic05: nit: I would enable the intellij option to flag variables that can be `final`, like these two

cryptography/hedera-cryptography-altbn128/src/main/java/com/hedera/cryptography/altbn128/facade/Group2.java
```diff
@@ -0,0 +1,268 @@
...
+
+    /**
+     * Creates a Group2 point from a random seed.
+     * @param seed the byte array seed.
+     * @return the byte array representation of the Group2 point.
+     * @throws AltBn128Exception in case of error.
+     */
+    public byte[] fromSeed(@NonNull final byte[] seed) {
+        validateSize(seed, randomSeedSize, "Invalid random seed size");
+        final ByteBuffer output = ByteBuffer.allocate(size);
```
lpetrovic05: any reason to use a `ByteBuffer` instead of just an array? if yes, its worth adding a comment explaining it

cryptography/hedera-cryptography-altbn128/src/main/java/com/hedera/cryptography/altbn128/AltBn128Group2.java
edward-swirldslabs: @mxtartaglia-sl would it be correct to think of Group2 as the 2nd dimension in a 2D coordinate system?    Group2 is the same space as Group, but the former is the y-coordinate and the latter is the x-coordinate in a (x,y) pair?

cryptography/hedera-cryptography-altbn128/src/main/java/com/hedera/cryptography/altbn128/AltBn128Group2Element.java
```diff
@@ -0,0 +1,143 @@
...
+    @Override
+    public GroupElement compress() {
+        return this.copy();
+    }
+
+    /**
+     * {@inheritDoc}
+     * @deprecated
+     */
+    @Deprecated
```
edward-swirldslabs, line 103: Since this is new code, why do we have deprecated methods?

cryptography/hedera-cryptography-altbn128/src/main/java/com/hedera/cryptography/altbn128/common/BigIntegerUtils.java
```diff
@@ -0,0 +1,138 @@
...
+    /**
+     * Converts a little-endian byte array into a BigInteger.
+     *
+     * @param littleEndianBytes the byte array in little-endian order
+     * @return the corresponding BigInteger
+     */
+    @NonNull
+    public static BigInteger fromLittleEndianBytes(@NonNull final byte[] littleEndianBytes) {
+        Objects.requireNonNull(littleEndianBytes, "littleEndianBytes must not be null");
+        byte[] bigEndianBytes = reverseBytes(Arrays.copyOf(littleEndianBytes, littleEndianBytes.length));
```
edward-swirldslabs: Would it be faster to create a new array in the `reverseBytes` method and then copying the values in reverse order instead of copying the array with the same values and then swapping?   (Speed difference being one less copy of data occurring.)

cryptography/hedera-cryptography-altbn128/src/main/java/com/hedera/cryptography/altbn128/facade/Group2.java
```diff
@@ -0,0 +1,268 @@
...
+     *
+     * @param scalars a byte matrix representing a list of N scalars of {@link FieldsLibraryAdapter#fieldElementsSize()}  representing each scalar
+     * @return N points each as a byte array of {@link Group2LibraryAdapter#g2Size()} containing representation of the resulting point.
+     * @throws NullPointerException if scalars is null
+     * @throws AltBn128Exception in case of an error during the batch addition
+     */
+    public byte[][] batchMultiply(final byte[][] scalars) {
+        Objects.requireNonNull(scalars, "scalars must not be null");
+        final byte[][] array = new byte[scalars.length][this.size];
+        int result = adapter.g2batchScalarMultiplication(scalars, array);
```
edward-swirldslabs: can be final

edward-swirldslabs (dismissed):

cryptography/hedera-cryptography-altbn128/src/main/java/com/hedera/cryptography/altbn128/AltBn128Group2.java
mxtartaglia-sl: Ok so, first thing first, the javadoc is wrong; it should be `{@link Group}`

Also, I changed the `com.hedera.cryptography.altbn128.facade.Group2` to `com.hedera.cryptography.altbn128.facade.Group2Facade` so it is less confusing.

A group is a set of points of a predefined curve; any operations on those points produce points on the same set. Two groups define Alt-bn-128. Group1 points coordinates (x,y) are slightly shorter than group2 coordinates. 
A bilinear pairing is a map of a point in group1 and a point in group2 to a point in another group we can ignore. In that sense, you can think  (pg1, pg2) as a pair we can verify against another pair, with a function e.
We chose to use group2 as the public key group, and group1 as the signature group, which means that the 
pair of points e(signature, publickey) can be validated.

image
The `Group` interface corresponds to a mathematical abstraction. It allows us to get some known points (zero and generator), deserialize, and perform some operations (batch add, batch multiply), as well as instantiate points (`GroupElements`) that belong to that group. All groups expose the same behavior, which is why they implement the same `Group` interface; it is just that the points produced by `Group1` and `Group2` are different (their coordinates have different scalar fields). Adaptors must be specific to each group, given rust limitations.

I hope it shines some light, but if not, we can address it in a call.

cryptography/hedera-cryptography-altbn128/src/main/java/com/hedera/cryptography/altbn128/AltBn128Group2.java
mxtartaglia-sl: I thought i did, but i keep finding ways to forget

cryptography/hedera-cryptography-altbn128/src/main/java/com/hedera/cryptography/altbn128/facade/Group2.java
mxtartaglia-sl: changed to byte[]. Initially, I wanted the possibility to flip buffers if needed, but that did not work as I expected.

cryptography/hedera-cryptography-altbn128/src/main/java/com/hedera/cryptography/altbn128/AltBn128Group2Element.java
mxtartaglia-sl: Please take a look at:
https://github.com/hashgraph/hedera-services/issues/15120
I did not merge all those changes in this pr to limit the number of changes.
I will also update the proposal reflecting those changes, and some others related to serialization at a later point.

cryptography/hedera-cryptography-altbn128/src/main/java/com/hedera/cryptography/altbn128/common/BigIntegerUtils.java
mxtartaglia-sl: it would, thanks

cryptography/hedera-cryptography-altbn128/src/main/java/com/hedera/cryptography/altbn128/AltBn128Group2.java
lpetrovic05: thank you

cryptography/hedera-cryptography-altbn128/src/main/java/com/hedera/cryptography/altbn128/AltBn128Group2.java
```diff
@@ -0,0 +1,176 @@
...
+        final byte[][] all = new byte[elems.size()][];
+        for (int i = 0; i < elems.size(); i++) {
+            all[i] = elems.get(i).toBytes();
+        }
+        final byte[][] g2Elements = facade.batchMultiply(all);
+
+        return Arrays.stream(g2Elements)
+                .map(rep -> (GroupElement) new AltBn128Group2Element(this, rep))
+                .toList();
+    }
```
lpetrovic05: there is a lot of duplication here that can me extracted into a method I think. also, one method uses `getInnerRepresentation` while the other uses `toBytes`, which is the same data, but being copied in one method and not the other. any reason for the discrepancy?

mxtartaglia-sl: You are right; the reason why groupElements have a innerRepresentation and the fieldElements don't is that I was preparing for two serialization mechanisms for groupElements, toBytes would return the "external" serialization (an affine representation of the point aka (x,y) pair), while innerRepresentation would return one that did not need to transform to affine (the projective)
Then, I discovered that both have the same serialization. 

The other problem we are covering is avoiding someone messing with the internals of the byte[] once exposed; that is why toBytes copies the array.

We can use the same approach for fieldElements, as the fieldElements operations have not been completed and will be addressed here: https://github.com/hashgraph/hedera-services/issues/14786.

cryptography/hedera-cryptography-altbn128/src/main/java/com/hedera/cryptography/altbn128/adapter/Group2LibraryAdapter.java
mxtartaglia-sl: called it `POINT_COORDINATE_BYTE_SIZE` and added the javadoc

cryptography/hedera-cryptography-altbn128/src/main/java/com/hedera/cryptography/altbn128/AltBn128Group2.java
mxtartaglia-sl: There, made `groupElement` and `fieldElement` consistent

edward-swirldslabs (approved):

PR #46, opened 2024-08-27
chore: 15120 renames and reorg

### LINKED ISSUE #15120: Pairings Api Name refactors
Given the recent confusion in trying to explain the concepts, i came up with a new organization and naming for some of the pieces.

mxtartaglia-sl:
**Description:**
Given some of the challenges and confusion related to naming, I am providing a different naming strategy.

image

Related issue(s):

Fixes https://github.com/hashgraph/hedera-services/issues/15120

**Notes for reviewer:**
Binary files are provided until compiling support with gradle for rust is completed.


lpetrovic05:
the diff looks like its a lot more than a rename

mxtartaglia-sl:
@lpetrovic05 this depends on the previous pr, i changed the baseline to reflect that.
Thanks

cryptography/hedera-cryptography-altbn128/src/main/java/module-info.java
```diff
@@ -6,9 +8,9 @@
     requires com.hedera.cryptography.pairings.api;
     requires jdk.jshell;
 
-    uses com.hedera.cryptography.pairings.spi.BilinearPairingProvider;
+    uses PairingFriendlyCurveProvider;
```
edward-swirldslabs, line 11: Is it better to have the full package path on this line, or in the import statement at the top of the file?  It seems to be standard practice to not use import statements in module-info?

edward-swirldslabs (dismissed):

cryptography/hedera-cryptography-altbn128/src/main/java/module-info.java
mxtartaglia-sl: This is IntelliJ doing when refactoring. I keep coming back and changing it back to the complete reference. 
I think better let's go with what intellij does by default.

edward-swirldslabs: ah, I wonder if that is a configurable setting.   Agreed to just adopt what intellij does or have spotless apply fix it if it is a coding style error.

lpetrovic05 (approved):

cryptography/hedera-cryptography-altbn128/src/main/java/module-info.java
mxtartaglia-sl: Resolving this, i'll take a look later if it can be modified.

PR #61, opened 2024-09-05
feat: 15024 Private key Sign method

### LINKED ISSUE #15024: `pairings-signatures-library`: Create the method that signs a byte array
Following: https://github.com/hashgraph/hedera-services/blob/develop/platform-sdk/docs/proposals/TSS-Library/TSS-Library.md
Section: #Hedera Cryptography Pairings Signatures Library
Implement the Signature class and possibility of signing a byte array message using a private key
- sign a message: It is the scalar multiplication of the private key element, and the `hashToGroup` `GroupElement` from the message, of the `Group` configured in the `SignatureSchema` for signatures.
- serialization of the signature.

The creation of the signature should be responsibility of the private key class
The serialization of the signature is responsibility of the signature class.

Note on serialization, it should be produced with ByteString class of protobuf. The first byte should include the serialized form of the SignatureSchema

Depends on: 
[#14788](https://github.com/hashgraph/hedera-services/issues/14788)
[#14786](https://github.com/hashgraph/hedera-services/issues/14786)

mxtartaglia-sl:
**Description**:
Exposing operation to sign a message

**Related issue(s)**:

Fixes [#15024](https://github.com/hashgraph/hedera-services/issues/15024)

**Notes for reviewer**:
Binary files are provided until compiling support with Gradle for Rust is completed.

**Checklist**

### DIFF
--- a/cryptography/hedera-cryptography-pairings-signatures/src/main/java/com/hedera/cryptography/pairings/signatures/api/ByteArrayConversionUtils.java
+++ b/cryptography/hedera-cryptography-pairings-signatures/src/main/java/com/hedera/cryptography/pairings/signatures/api/ByteArrayConversionUtils.java
@@ -1,0 +1,148 @@
+/*
+ * Copyright (C) 2024 Hedera Hashgraph, LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.hedera.cryptography.pairings.signatures.api;
+
+import com.hedera.cryptography.pairings.api.PairingsException;
+import edu.umd.cs.findbugs.annotations.NonNull;
+import java.nio.ByteBuffer;
+import java.util.Arrays;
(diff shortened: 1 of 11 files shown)


edward-swirldslabs (dismissed):

cryptography/hedera-cryptography-pairings-signatures/src/main/java/com/hedera/cryptography/pairings/signatures/api/ByteArrayConversionUtils.java
```diff
@@ -0,0 +1,148 @@
...
+class ByteArrayConversionUtils {
+
+    /**
+     * Returns a {@link PairingPrivateKey} instance out of this object serialized form
+     * @param bytes the serialized form of this object
+     * @return a {@link PairingPrivateKey} instance
+     * @throws IllegalArgumentException if the key representation is invalid
+     */
+    @NonNull
+    static PairingPrivateKey deserializePairingPrivateKey(@NonNull final byte[] bytes) {
```
edward-swirldslabs, line 40: Given our discussion earlier today, should this be bytes or a more detailed serialized form?

edward-swirldslabs (dismissed):

cryptography/hedera-cryptography-pairings-signatures/src/main/java/com/hedera/cryptography/pairings/signatures/api/ByteArrayConversionUtils.java
mxtartaglia-sl: I am keeping the byte array strategy for now.
I am still unclear on the benefit of using a different alternative or the challenges that this strategy poses for consumers.
Each key or signature can return the Paring element, which can be converted to BigInteger coordinates if we wish to; the only difference would be that we delete this method or add a shortcut to that accessor.

edward-swirldslabs (approved):

PR #70, opened 2024-09-19
build: rust compilation support

jjohannes:
**Description**:
Integrate Gradle and Rust/Cargo:

- [x] Gradle tasks for Rust compilation
- [x] Install Rust toolchain via GitHub Actions
- [x] Build the `linux-x86-64` target
- [x] Enhance Rust toolchain setup to support cross-compilation for all desired targets
   - Windows cross-compilation using `lld` based on https://jake-shadle.github.io/xwin/ 

**Related issue(s)**:

https://github.com/hashgraph/hedera-services/issues/14100

**Notes for reviewer**:

**Checklist**


mxtartaglia-sl (dismissed):
This is great, Jendrik! With this implementation, I can stop uploading the binaries for unit test to pass.
The only comment I would make is that maybe it is worth having the targets in some sort of enum, but I didn't investigate whether that is possible.

mxtartaglia-sl (approved):

mxtartaglia-sl (dismissed):

edward-swirldslabs (dismissed):

mxtartaglia-sl (dismissed):

jjohannes:
Once this is in, we could extend the pipeline to run all tests on runners with different operating systems. The built libraries are pulled from the Gradle cache. I have already tested that on a branch (#82) and [it works](https://github.com/hashgraph/hedera-cryptography/actions/runs/11368312816?pr=82). 

Discovered this issue: https://github.com/hashgraph/hedera-cryptography/issues/81

Prepared follow up PR (for once this is merged): #82

gradle/plugins/src/main/kotlin/com.hedera.gradle.feature.rust.gradle.kts
```diff
@@ -0,0 +1,35 @@
...
+import com.hedera.gradle.services.TaskLockService
+import com.hedera.gradle.tasks.CargoBuildTask
+
+plugins { id("java") }
+
+val cargo = project.extensions.create<CargoExtension>("cargo")
+
+cargo.targets(aarch64Darwin, aarch64Linux, x86Darwin, x86Linux, x86Windows)
+
+// Cargo might do installation work, do not run in parallel:
```
rbarker-dev, line 28: why not run in parallel?

rbarker-dev (approved):

README.md
rbarker-dev, line 1: To get `./gradlew build` to succeed I also needed to execute the following command: `rustup target add x86_64-pc-windows-msvc`

rbarker-dev: I think we should put that in as a requirement for building on mac.

rbarker-dev (changes requested):

README.md
jjohannes: Good catch! Added it to the Readme.

gradle/plugins/src/main/kotlin/com.hedera.gradle.feature.rust.gradle.kts
jjohannes: Cargo/rust is doing installation/preparation work before compiling our code. If we run this in parallel, we start `cargo` multiple times in parallel. It then tries to do the same things multiple times in parallel and fails with files access issues (it's apparently not designed to run in parallel).
I tried to isolate this into [a share build service](https://docs.gradle.org/current/userguide/build_services.html), but did not succeed as this is triggered automatically when you start the (cross) compilation through cargo. We would somehow need to make sure that everything is prepared before the actual compilation starts. But I did not get it working (maybe cargo does support that at all).

rbarker-dev (approved):
LGTM

mxtartaglia-sl (approved):

mxtartaglia-sl:
Tested locally and working. Both in Linux and in mac.

PR #71, opened 2024-09-19
15201: Equals and Hashcode handled by java

mxtartaglia-sl:
**Description**:
This pr aims to provide an optimal implementation of hashcode and equal methods

Fixes #[#15201](https://github.com/hashgraph/hedera-services/issues/15201)

**Notes for reviewer**:

**Checklist**


jjohannes (dismissed):

cryptography/hedera-cryptography-altbn128/src/jmh/java/com/hedera/cryptography/altbn128/FieldElementEqualityAndHashCodeBenchmark.java
```diff
@@ -0,0 +1,98 @@
...
+import org.openjdk.jmh.annotations.Scope;
+import org.openjdk.jmh.annotations.Setup;
+import org.openjdk.jmh.annotations.State;
+import org.openjdk.jmh.annotations.Threads;
+import org.openjdk.jmh.annotations.Warmup;
+import org.openjdk.jmh.infra.Blackhole;
+
+@State(Scope.Benchmark)
+public class FieldElementEqualityAndHashCodeBenchmark {
+    private List<Long> values;
```
lpetrovic05: seems this is unused

lpetrovic05:
I was looking at `AltBn128FieldElement.hashCode()`. Atm it is:
`Objects.hash(this.getClass(), this.field, Arrays.hashCode(this.representation));`
But I think it can be simplified to just: `Arrays.hashCode(this.representation)` because `this.getClass()` adds no additional value, since its the same class for the same types. And `field` is similar, it is just a link to the native API. 
I could not run the JMH test to verify what the impact of this is.

lpetrovic05:
I also looked at the `equals` method:
```
    public boolean equals(final Object obj) {
        if (!(obj instanceof AltBn128FieldElement)) return false;
        if (this.field != ((AltBn128FieldElement) obj).field) return false;
        if (this.representation.length != ((AltBn128FieldElement) obj).representation.length) return false;

        return Arrays.equals(this.representation, ((AltBn128FieldElement) obj).representation);
    }
```

Why are we checking if `field` is the same instance?

cryptography/hedera-cryptography-altbn128/src/jmh/java/com/hedera/cryptography/altbn128/FieldElementEqualityAndHashCodeBenchmark.java
mxtartaglia-sl: Will check! thanks

mxtartaglia-sl:
> Why are we checking if the field is the same instance?

Maybe this is a bit of a hack, but the field is used as a singleton; we are not creating instances of Field class throughout the code; it is almost a static factory. Because of that, it does not define an equals or hashCode method. So, it inherits that from the object, which is the same instance comparison. I am open to suggestions.

> But I think it can be simplified to just: Arrays.hashCode(this.representation) because this.getClass() adds no additional value ...  And field is similar, it is just a link to the native API.

Erring on the side of caution. I wanted to decrease the chance that two instances belonging to different curve implementations could return the same hashCode because their internal representation is the same. The inner representation array is from one to four numbers. I imagine future curve implementations with the same field elements and/or points. We would want the equals method to return false (for same numeric value points belonging to different curves), and I think we would also want the hashcode to be different.

rbarker-dev (dismissed):

lpetrovic05:
> > Why are we checking if the field is the same instance?
> 
> Maybe this is a bit of a hack, but the field is used as a singleton; we are not creating instances of Field class throughout the code; it is almost a static factory. Because of that, it does not define an equals or hashCode method. So, it inherits that from the object, which is the same instance comparison. I am open to suggestions.

If its supposed to be a singleton, then there is no need to check. If there is some kind of error and we have two instances of it, I don't think we want the equals to fail.

> > But I think it can be simplified to just: Arrays.hashCode(this.representation) because this.getClass() adds no additional value ...  And field is similar, it is just a link to the native API.
> 
> Erring on the side of caution. I wanted to decrease the chance that two instances belonging to different curve implementations could return the same hashCode because their internal representation is the same. The inner representation array is from one to four numbers. I imagine future curve implementations with the same field elements and/or points. We would want the equals method to return false (for same numeric value points belonging to different curves), and I think we would also want the hashcode to be different.

Two instances having the same hash code is not a huge issue, this is expected to happen, the worst thing it can do is lead to inefficiency, `equals` is always used as a fall back. So I think we should just use the array.

mxtartaglia-sl:
> Two instances having the same hash code is not a huge issue, this is expected to happen, the worst thing it can do is lead to inefficiency, equals is always used as a fall back. S

Why would we want to simplify to force the inefficiency? What would be the benefit?

lpetrovic05:
> > Two instances having the same hash code is not a huge issue, this is expected to happen, the worst thing it can do is lead to inefficiency, equals is always used as a fall back. S
> 
> Why would we want to simplify to force the inefficiency? What would be the benefit?

I don't think it would be more inefficient, I just think that is the worst case scenario. It is probably more efficient to just use the byte array if I were to guess.

lpetrovic05 (dismissed):

lpetrovic05 (approved):

rbarker-dev (approved):

PR #78, opened 2024-10-08
build: use common Hiero Gradle config

jjohannes:
**Description**:

Before this can be integrated:

- [x] Integrate #70 
- [x] Publish _hiero-gradle-conventions_ plugins to Gradle Plugin Portal and use from there

**Related issue(s)**:

https://github.com/hashgraph/hedera-services/issues/14255

**Notes for reviewer**:

**Checklist**


.github/workflows/zxc-release-maven-central.yaml
```diff
@@ -159,12 +144,18 @@ jobs:
...
-      - name: Gradle Publish to Maven Central
+      # Publishes the artifacts to the Maven Central Nexus staging repository.
+      # TODO for initial release and testing, we use -PpublishTestRelease to not automatically release from staging.
+      - name: Gradle Publish to Maven Central - common
+        if: ${{ inputs.dry-run-enabled != true && !cancelled() && !failure() }}
+        env:
+          NEXUS_USERNAME: ${{ secrets.ossrh-user-name }}
+          NEXUS_PASSWORD: ${{ secrets.ossrh-user-password }}
+        run: ./gradlew releaseMavenCentral -PpublishingPackageGroup=com.hedera.common -PpublishSigningEnabled=true --no-configuration-cache --scan -PpublishTestRelease=true
+
```
jjohannes, line 155: Publishing is split into two steps for the different groups `com.hedera.common` and `com.hedera.cryptography`. Technically, we can publish both at once with the `com.hedera` base group, but I would like to keep things consistent across repositories and we do not have a common "base" group everywehre.

`-PpublishingPackageGroup` now has to match exactly the group of the projects. This makes the Gradle configuration no longer rely on hard coded "base" group values.

There is a good amount of automated testing for this setup now that all repositories will share in the future:
https://github.com/hiero-ledger/hiero-gradle-conventions/blob/main/src/test/kotlin/org/hiero/gradle/test/MavenCentralPublishTest.kt

.github/workflows/build.yml
```diff
@@ -58,14 +59,11 @@ jobs:
       - name: Build software
         run: ./gradlew assemble --scan
 
-      - name: Spotless Check
-        run: ./gradlew spotlessCheck --scan --continue
-
-      - name: Dependency Scopes Check
-        run: ./gradlew checkAllModuleInfo --scan --continue
+      - name: Code Quality Checks
+        run: ./gradlew qualityCheck --continue --scan
```
jjohannes, line 63: All projects now have the `qualityCheck` task which does the same as `qualityGate` but without auto-fixing.
I suggest to use this step in all CI pipelines instead of splitting it into multiple "check steps".
This way, we control the whole set of checks in the Gradle plugins and can add a new check to it – like the recently added POM validation – without making changes to all CI pipelines.

.github/dependabot.yml
```diff
@@ -1,5 +1,11 @@
 version: 2
 updates:
+  - package-ecosystem: gradle
+    directory: "/hiero-dependency-versions"
+    schedule:
+      interval: weekly
+    open-pull-requests-limit: 10
+
```
jjohannes, line 8: This was missing. I don't know why Dependabot does not find this subfolder automatically, but it doesn't. That's why we never got PRs for library versions in this repository (we only got PRs for Gradle plugin versions).

mxtartaglia-sl (dismissed):

mxtartaglia-sl (approved):

rbarker-dev (approved):

PR #79, opened 2024-10-10
feat: KeyGen should use signatures library

### LINKED ISSUE #15199: Refactor Ec-Key Gen Tool to use `pairings-signatures-library`
The ec-key-gen tool (compiled in java) currently invokes arkworks without the `pairings-api`. The tool must be updated to use the `pairings-signatures-library` instead of arkworks through jni.

The tool should then remain simple and create a file based on the base64 translation of the bytes returned by the signature library.

Any upgrade in the interface is welcomed as no users are still affected by this change.
The tool added three padding bytes, which should be discarded as they are no longer necessary.

lpetrovic05:
closes https://github.com/hashgraph/hedera-services/issues/15199


cryptography/hedera-cryptography-eckeygen/src/main/java/com/hedera/cryptography/eckeygen/KeyGen.java
```diff
@@ -48,80 +54,99 @@ private KeyGen() {
...
+    }
+
+    /**
+     * Generates a Key Pair (private and public keys)
+     *
+     * @return a key pair
+     * @throws NoSuchAlgorithmException if no algorithm found to get a {@link SecureRandom} instance
+     */
+    @NonNull
+    public static PairingKeyPair generateKeyPair() throws NoSuchAlgorithmException {
```
mxtartaglia-sl: Maybe we can move this responsibility to the `PairingKeyPair` class? Otherwise, it is just a container.

mxtartaglia-sl (dismissed):

cryptography/hedera-cryptography-eckeygen/src/main/java/com/hedera/cryptography/eckeygen/KeyGen.java
lpetrovic05: good idea, done

rbarker-dev (approved):
Review covers
- `cryptography/hedera-cryptography-pairings-signatures/build.gradle.kts`

Looks good.

mxtartaglia-sl (approved):

Issue #81, opened 2024-10-16
Rust: Windows DLL loading is not using correct file name


jjohannes:
### Description

The DLL compiled for Windows does not have the `lib` prefix, which means for the library `bn254` the file is called `bn254.dll` and **not** `libbn254.dll`.

We could probably rename the file as part of the build if we want that solution, but maybe it is more consistent with common standards to not prepend the "lib" prefix for Windows in the loading code.

### Steps to reproduce

I created a setup that runs the `hedera-cryptography-altbn128` tests on Windows on CI on top of #70.
It's failing to load the file, although it was created.
- https://github.com/hashgraph/hedera-cryptography/actions/runs/11363866764/job/31608808094
- https://scans.gradle.com/s/q23d7xbiz4cc2/tests/overview

### Version

develop

### Operating system

Windows

jjohannes:
I prepared a PR for running the tests on CI on Windows/Mac.

If you start working on this, you may use that to test if you do not have a local Windows environment.

https://github.com/hashgraph/hedera-cryptography/pull/82

mxtartaglia-sl:
@timo0 It would be very good feedback how you find interacting with the build process @jjohannes was working.
As I have most of the necessary things already installed.

mxtartaglia-sl:
@poulok this one is in hedera-cryptography repo, I don't know if its gonna show up in burncharts and your reports

jjohannes:
Anther small thing: https://github.com/hashgraph/hedera-cryptography/pull/82#discussion_r1804948461
Added a fix to #82.

PR #82, opened 2024-10-16
Run hedera-cryptography-altbn128 tests on multiple operating systems

jjohannes:
**Description**:

- Follow up to #70 
- Affected by #81


jjohannes:
The CI build sometimes runs into this problem with `xwin`: https://github.com/Jake-Shadle/xwin/issues/141

If it fails, trigger a re-run. I hope there will be a fix or stable workaround soon.

gradle/plugins/src/main/kotlin/com/hedera/gradle/tasks/CargoBuildTask.kt
```diff
@@ -85,11 +85,13 @@ abstract class CargoBuildTask : DefaultTask() {
 
     private fun installCargoCrossCompiler(buildsForWindows: Boolean) {
         exec.exec {
-            val crossCompiler = if (buildsForWindows) "xwin" else "cargo-zigbuild"
+            val crossCompiler = if (buildsForWindows) "xwin@0.6.6-rc.1" else "cargo-zigbuild"
```
rbarker-dev: Can you add a comment linking to the thread where it was determined we needed to use 0.6.6-rc.1 here in the code? That way if a more stable version is released it should be a little easier to justify switching to it.

rbarker-dev (changes requested):

gradle/plugins/src/main/kotlin/com/hedera/gradle/tasks/CargoBuildTask.kt
jjohannes: :+1: Comment added. I am following the _xwin_ issue and will keep an eye on this to remove it again once it is no longer needed.

common/hedera-common-nativesupport/src/main/java/com/hedera/common/nativesupport/NativeLibrary.java
```diff
@@ -252,7 +252,10 @@ private void installUnchecked(@NonNull final Class<?> c) {
     @SuppressWarnings("ResultOfMethodCallIgnored")
     private void install(@NonNull final InputStream resourceStream) throws IOException {
         Objects.requireNonNull(resourceStream, "resourceStream must not be null");
-        final String fileName = Path.of(name).getFileName().toString();
+        final OperatingSystem os = OperatingSystem.current();
+        final String libName =
+                os == OperatingSystem.WINDOWS ? name + "." + libExtensions.get(OperatingSystem.WINDOWS) : name;
+        final String fileName = Path.of(libName).getFileName().toString();
```
jjohannes, line 258: @timo0 now that the tests are running on Windows, I found this other special case: On Windows, the DLL extension is part of the library name to load.

On the upside, with this change the tests pass on Windows. Which is a good sign that the cross-compilation is successful. 🎉

mxtartaglia-sl: thanks @jjohannes

mxtartaglia-sl (dismissed):

.github/workflows/build.yml
```diff
@@ -82,5 +82,41 @@ jobs:
       - name: Publish Unit Test Results
         uses: step-security/publish-unit-test-result-action@4519d7c9f71dd765f8bbb98626268780f23bab28 # v2.17.0
         with:
-          files: |
-            test-results/**/*.xml:
+          files: "**/build/test-results/test/TEST-*.xml"
```
jjohannes, line 85: Not related directly, but the path here was wrong and thus test results were not published to PRs.

mxtartaglia-sl (approved):

.github/workflows/build.yml
mxtartaglia-sl: thanks!

rbarker-dev (approved):
LGTM

PR #83, opened 2024-10-16
fix: Respect prefixes for different OS

### LINKED ISSUE #81: Rust: Windows DLL loading is not using correct file name

timo0:
Fix #81


common/hedera-common-nativesupport/src/main/java/com/hedera/common/nativesupport/NativeLibrary.java
```diff
@@ -155,12 +168,20 @@ public String locationInJar() {
...
         Objects.requireNonNull(libraryName, "name must not be null");
+        Objects.requireNonNull(libPrefixes, "libPrefixes must not be null");
         Objects.requireNonNull(libExtensions, "libExtensions must not be null");
         final OperatingSystem os = OperatingSystem.current();
         final Architecture arch = Architecture.current();
 
+        String libPrefix = libPrefixes.get(os);
+        if (libPrefix == null) {
+            libPrefix = "";
+        }
```
mxtartaglia-sl: MAybe a.getOrDefault() to avoid some lines of code.
```suggestion
        final String libPrefix = libPrefixes.getOrDefault(os,"");
```

common/hedera-common-nativesupport/src/test/java/com/hedera/common/nativesupport/NativeLibraryTest.java
```diff
@@ -50,7 +50,7 @@ private static Stream<Object[]> combinedParameters() {
     void testWithNameAndCustomExtensions(OperatingSystem operatingSystem, Architecture architecture) {
         final Map<OperatingSystem, String> extensions =
                 Map.of(OperatingSystem.WINDOWS, "win", OperatingSystem.LINUX, "lx", OperatingSystem.DARWIN, "dar");
-        NativeLibrary library = NativeLibrary.withName("libcustom", extensions);
+        NativeLibrary library = NativeLibrary.withName("custom", extensions, extensions);
```
mxtartaglia-sl: For this case I would not reuse the same map. Just to make sure that whatever was sent as extension was  used as extension, and what was sent as prefix was used as prefix

common/hedera-common-nativesupport/src/test/java/com/hedera/common/nativesupport/NativeLibraryTest.java
```diff
@@ -128,14 +129,14 @@ public void testInstallExistentLib() {
 
     @Test
```
mxtartaglia-sl, line 153: Can you add a test case scenario where it is informed only to a specific `os`, an check that for that os the prefix comes informed and for the rest is blank?

common/hedera-common-nativesupport/src/main/java/com/hedera/common/nativesupport/NativeLibrary.java
```diff
@@ -83,11 +83,18 @@ public class NativeLibrary {
     private static final Map<OperatingSystem, String> DEFAULT_LIB_EXTENSIONS =
             Map.of(OperatingSystem.WINDOWS, "dll", OperatingSystem.LINUX, "so", OperatingSystem.DARWIN, "dylib");
 
+    /**
+     * Default prefix for binary libraries per OS
+     */
+    private static final Map<OperatingSystem, String> DEFAULT_LIB_PREFIXES =
+            Map.of(OperatingSystem.WINDOWS, "", OperatingSystem.LINUX, "lib", OperatingSystem.DARWIN, "lib");
```
mxtartaglia-sl, line 90: We would only save some bits that would always be in memory,  but if we have an if statement that takes care of this case, we might as well not set the empty value for windows. ..

common/hedera-common-nativesupport/src/test/java/com/hedera/common/nativesupport/NativeLibraryTest.java
```diff
@@ -102,7 +103,7 @@ void testWithNameAndEmptyExtensions(OperatingSystem operatingSystem, Architectur
     void testWithNameAndBlankExtensions(OperatingSystem operatingSystem, Architecture architecture) {
         final Map<OperatingSystem, String> extensions =
                 Map.of(OperatingSystem.WINDOWS, "", OperatingSystem.LINUX, "", OperatingSystem.DARWIN, "");
-        NativeLibrary library = NativeLibrary.withName("libcustom", extensions);
+        NativeLibrary library = NativeLibrary.withName("custom", extensions, extensions);
```
mxtartaglia-sl: The name `extensions` is not accurate anymore if we are reusing the map

common/hedera-common-nativesupport/src/main/java/com/hedera/common/nativesupport/NativeLibrary.java
timo0: done :)

common/hedera-common-nativesupport/src/test/java/com/hedera/common/nativesupport/NativeLibraryTest.java
timo0: done

common/hedera-common-nativesupport/src/test/java/com/hedera/common/nativesupport/NativeLibraryTest.java
timo0: Added `testWithOnlyOneOsPrefix`

common/hedera-common-nativesupport/src/main/java/com/hedera/common/nativesupport/NativeLibrary.java
timo0: Sorry disagree with that this is more explicit and would like to keep that

common/hedera-common-nativesupport/src/test/java/com/hedera/common/nativesupport/NativeLibraryTest.java
timo0: done

common/hedera-common-nativesupport/src/main/java/com/hedera/common/nativesupport/NativeLibrary.java
mxtartaglia-sl: Makes sense

mxtartaglia-sl (approved):

PR #86, opened 2024-10-17
feat: 15211 create and decrypt elGamal cipher texts

### LINKED ISSUE #15211: `pairings-signatures-library`: Create ElGamalUtil Class to with the ability to create and decrypt a valid ElGamal cipher text
Create a utility class that can create and read ElGamal ciphertexts. 

ElGamalCipherText consists of two elements `(r, s),` where `r` is a random value and `s` is computed from the message, an encryption key, and `r.`

Signatures:
```
  ElGamalCipherText createCipherText(PairingsPublicKey key, FieldElement value);
  FieldElent readCipherText(PairingsPrivateKey key, Map<GroupElement, FieldElement> substitutionTable, ElGamalCipherText cipher);
```

`readCipherText` method will use a cipher previously calculated inversed substitution table for each possible value in a byte to the result of calculating `value * keyGroup.generator()`

mxtartaglia-sl:
**Description**:
Create a utility class that can create and read ElGamal ciphertexts.

**Related issue(s)**:

Fixes [#15211](https://github.com/hashgraph/hedera-services/issues/15211)


cryptography/hedera-cryptography-bls/src/main/java/com/hedera/cryptography/bls/extensions/elGamal/ElGamalUtils.java
```diff
@@ -0,0 +1,215 @@
...
+                // There are two possible approaches to handle this situation:
+                // 1. Return a random value to ensure the process remains consistent,
+                //    even though the decryption was unsuccessful.
+                // 2. Fail fast by throwing an exception.
+                //
+                // The issue with the second approach (throwing an exception) is that in some cases,
+                // even with an invalid decryption key, we may still produce valid commitments.
+                // If users rely solely on the exception to detect decryption failure, they might be misled,
+                // since the absence of an exception does not always indicate successful decryption.
+                value = (byte) (i % Byte.MAX_VALUE);
```
mxtartaglia-sl: @rsinha Can i get your feedback here?

mxtartaglia-sl: As we discussed on the call with @rsinha , we conveyed that the approach was faulty.
Using this approach, we could have a very specific scenario in which a valid message can be decrypted with an invalid key.
Correctness is only guaranteed for honest decryption attempts. It is okay to have an inconsistent result when trying a dishonest decrypt attempt or mistakenly providing the wrong input. So, in those cases, when no value can be obtained from the substitution table, a null value is returned as decryption result. If otherwise, we find a value, then we return a result of the decrypt that might not match the result.

lpetrovic05:
this seems to have the same set of changes as the other PR, should the target branch be changed?

cryptography/hedera-cryptography-tss/src/test/java/com/hedera/cryptography/tss/extensions/ElGamalUtilsTest.java
```diff
@@ -0,0 +1,160 @@
...
+        final int seed = INIT_RANDOM.nextInt();
+        System.out.println("Seed used: " + seed);
+        final Random random = new Random(seed);
+        final BlsPrivateKey sk = BlsPrivateKey.create(schema, random);
+        final BlsPublicKey pk = sk.createPublicKey();
+        final Map<Byte, FieldElement> substitutionTable = ElGamalUtils.elGamalSubstitutionTable(schema);
+
+        final var secret =
+                """
+                If you can't explain it simply, you don't understand it well enough.
```
edward-swirldslabs, line 139: To be fair, some subjects are intrinsically complex and divorced from ordinary common life experience.

edward-swirldslabs (approved):

PR #97, opened 2024-10-24
feat: 16103 `ElGamal` CipherTextTable and a CombinedCiphertext

### LINKED ISSUE #16103: `ElgamalUtils` Create an operation that can produce a multi-share ciphertext and a combined-ciphertext

mxtartaglia-sl:
**Description**:
Create an operation that can produce a CipherTextTable and a CombinedCiphertext

**Related issue(s)**:

Fixes [#16103](https://github.com/hashgraph/hedera-services/issues/16103)

**Checklist**


cryptography/hedera-cryptography-tss/src/main/java/com/hedera/cryptography/tss/extensions/elgamal/CiphertextTable.java
```diff
@@ -0,0 +1,62 @@
...
+import edu.umd.cs.findbugs.annotations.NonNull;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * A {@link CiphertextTable} contains an encrypted share from an individual secret, to each existing share.
+ *
+ * @param sharedRandomness a shared randomness for all the messages in {@code shareCiphertexts}
+ * @param shareCiphertexts the contained share ciphertexts, in order of the share IDs
```
lpetrovic05: A map is not necessarily ordered, only some implementations are. If ordering is important, then another structure might be better suited

cryptography/hedera-cryptography-tss/src/main/java/com/hedera/cryptography/tss/extensions/elgamal/ElGamalSubstitutionTable.java
```diff
@@ -0,0 +1,108 @@
...
+import com.hedera.cryptography.pairings.api.Group;
+import com.hedera.cryptography.pairings.api.GroupElement;
+import edu.umd.cs.findbugs.annotations.NonNull;
+import edu.umd.cs.findbugs.annotations.Nullable;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Objects;
+
+/**
+ * A substitution table to create and read CipherTexts
```
lpetrovic05, line 31: add a javadoc for the generic types

cryptography/hedera-cryptography-tss/src/main/java/com/hedera/cryptography/tss/extensions/elgamal/CiphertextTable.java
mxtartaglia-sl: Yes, I am aware of that. This comment is a left over from when it was a list. I will remove it.
The idea is that order is needed; it should come from outside during the iteration.

cryptography/hedera-cryptography-tss/src/main/java/com/hedera/cryptography/tss/extensions/elgamal/ElGamalUtils.java
```diff
@@ -197,13 +178,69 @@ public static Map<Byte, FieldElement> elGamalSubstitutionTable(@NonNull final Si
...
+        final List<FieldElement> randomness = ElGamalUtils.generateEntropy(
+                random, signatureSchema.getPairingFriendlyCurve().field().elementSize(), signatureSchema);
+
+        final List<GroupElement> chunkRandomness = new ArrayList<>();
+        for (final FieldElement randomElement : randomness) {
+            chunkRandomness.add(publicKeyGenerator.multiply(randomElement));
+        }
+
+        final Map<Integer, List<GroupElement>> multiEncryptedValues = new HashMap<>();
+        for (int i = 0; i < ids.size(); i++) {
```
edward-swirldslabs, line 234: It might be worth adding comments for blocks of code that do complex operations to explain what is happening at a higher level algorithmically.   Naively this looks like the construction of some or all of the contents of a TssMessage. <-- my uncertainty is what I am highlighting.   It will help future maintainers to be able to map our high level algorithm descriptions (TssMessage / Shares / Shares of Shares) to the low level mathematical implementation.

edward-swirldslabs (approved):

cryptography/hedera-cryptography-tss/src/main/java/com/hedera/cryptography/tss/extensions/elgamal/ElGamalUtils.java
mxtartaglia-sl: Ed, I understand the issue.
You are right; these classes are tools we will use to build the different parts of the tss message.
So here is the challenge: this is meant to be somewhat generic, so adding higher-level explanations about TSS specifics fell off here. I do agree, however, that this will become complicated in the future and that the efforts need to be explained.
I plan for the class and operation that coordinates all these calls to explain what we are doing, so in case somebody chooses to use these pieces for something else, the comments don't reflect an external intention.

edward-swirldslabs: Sounds good to me if it is being explained higher up.

PR #98, opened 2024-10-24
feat: 16060 Create Polynomial and FeldmanCommitment

### LINKED ISSUE #16060: Create Polynomial and FeldmanCommitment
Create a polynomial px  out of random values that can transport a secret FieldElement in x0 and can be evaluated at different points.
Create a Feldman commitment out of the polynomial by calculating GeneratorGroup * px. X0 of the commitment should correspond to the public part of the secret transmitted by the polynomial.

mxtartaglia-sl:
**Description**:
Create a polynomial px out of random values that can transport a secret FieldElement in x0 and can be evaluated at different points.
Create a Feldman commitment out of the polynomial by calculating GeneratorGroup * px. X0 of the commitment should correspond to the public part of the secret transmitted by the polynomial.
**Related issue(s)**:

Fixes https://github.com/hashgraph/hedera-services/issues/16060

**Notes for reviewer**:
Depends on other PRs
 https://github.com/hashgraph/hedera-cryptography/pull/97/files

**Checklist**

### DIFF
--- a/cryptography/hedera-cryptography-tss/src/main/java/com/hedera/cryptography/tss/extensions/FeldmanCommitment.java
+++ b/cryptography/hedera-cryptography-tss/src/main/java/com/hedera/cryptography/tss/extensions/FeldmanCommitment.java
@@ -70,14 +70,6 @@ public record FeldmanCommitment(@NonNull List<GroupElement> commitmentCoefficien
     public GroupElement evaluate(@NonNull final FieldElement x) {
 
         Objects.requireNonNull(x, "x must not be null");
-        //         for msg in dkg_messages.iter() {
-        //            // compute powers of share_id
-        //            let xs = (0..msg.commitment.len()).map(|i| share_id.pow([i as u64])).collect::<Vec<ShareId<G>>>();
-        //            let share_of_share_pubkey = msg.commitment.iter().zip(xs.iter()).fold(G::zero(), |acc, (&a_i,
-        // &x_i)| { acc + a_i.mul(x_i) });
-        //            dealt_share_pubkeys.push(share_of_share_pubkey);
-        //        }
-
         GroupElement result = commitmentCoefficients.getFirst().getGroup().zero();
         for (int i = 0; i < commitmentCoefficients.size(); i++) {
             final FieldElement exponentiatedShareId = x.power(i);
--- a/cryptography/hedera-cryptography-tss/src/main/java/com/hedera/cryptography/tss/extensions/FeldmanCommitment.java
+++ b/cryptography/hedera-cryptography-tss/src/main/java/com/hedera/cryptography/tss/extensions/FeldmanCommitment.java
@@ -1,0 +1,89 @@
+/*
+ * Copyright (C) 2024 Hedera Hashgraph, LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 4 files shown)


cryptography/hedera-cryptography-tss/src/main/java/com/hedera/cryptography/tss/extensions/FeldmanCommitment.java
```diff
@@ -0,0 +1,89 @@
...
+     * Evaluates the polynomial at a specific value without reveling the polynomial.
+     *
+     * @param x the value to evaluate the commitment
+     * @return a point on the curve
+     */
+    @NonNull
+    public GroupElement evaluate(@NonNull final FieldElement x) {
+
+        Objects.requireNonNull(x, "x must not be null");
+        //         for msg in dkg_messages.iter() {
```
edward-swirldslabs: What's left to decide about this commented code?

cryptography/hedera-cryptography-tss/src/test/java/com/hedera/cryptography/tss/extensions/PolynomialTest.java
```diff
@@ -0,0 +1,92 @@
...
+import com.hedera.cryptography.pairings.api.FieldElement;
+import com.hedera.cryptography.pairings.api.PairingFriendlyCurves;
+import java.security.SecureRandom;
+import java.util.List;
+import java.util.Random;
+import java.util.stream.IntStream;
+import org.junit.jupiter.api.Test;
+import org.mockito.Mockito;
+
+// FUTURE maybe use https://github.com/PoslavskySV/rings to match values against another implementation.
```
edward-swirldslabs: When does this comment become relevant?  Are there any negative implications now from this comment?

cryptography/hedera-cryptography-tss/src/main/java/com/hedera/cryptography/tss/extensions/FeldmanCommitment.java
mxtartaglia-sl: Thanks, I should remove this comment.

cryptography/hedera-cryptography-tss/src/test/java/com/hedera/cryptography/tss/extensions/PolynomialTest.java
mxtartaglia-sl: It is relevant for improving our ability to test this class.
Given that the polynomial is defined for a finite field, it is not possible to easily test it.
I think that if we add a library that can provide an alternative finite field implementation, we can calculate a reference polynomial with the second library I contrast the result of the two of them as a testing mechanism.
We could also consider alternatives to that, so I am open to suggestions.

edward-swirldslabs: Out of my depth. :-) @rsinha might better inform.    Is there anything like this test at the rust level that the core library uses for checking correctness?   If not, I wouldn't worry about it.  We're bought in on the soundness of the rust implementation.   Unless @rsinha feels this is needed and adds value.

edward-swirldslabs (approved):

cryptography/hedera-cryptography-tss/src/test/java/com/hedera/cryptography/tss/extensions/PolynomialTest.java
mxtartaglia-sl: I'll contact him. Given that we are separating the test part from the implementation, this comment will matter when we are in that part of the effort. 
Really, it depends on the pairings API. If the operations there work, then this one should work, too. There are not many moving parts to test.
In any case, if it is not necessary by then, I will remove the comment.
Thanks for the feedback!

edward-swirldslabs (dismissed):

edward-swirldslabs (approved):

PR #101, opened 2024-10-28
feat: modify hash to curve algorithm

### LINKED ISSUE #15451: Modify HashToCurve/HashToGroup algorithm
Messages (byte arrays) must be mapped or encoded to points in the curve to be signed.
The same hashing process/strategy must be followed where the signature is produced and in all places that verify that signature.

Our current algorithm produces a `SHA-256` hashed value of the message to sign, then gets a random curve point as if the hash was a random seed.

As explained by the S.C. team, cc: @Nana-EC  @AlfredoG87 and @david-bakin-sl, at least the hashing method should be changed to 'keccak256` (alternate to SHA-256) to ensure cross-chain compatibility. 

A POC implementation of the hashing mechanism in solidity can be observed here: https://github.com/hashgraph/hedera-smart-contracts/blob/main/contracts/bls-signature/Pairing.sol#L57-L59

```
   function hashToG1(bytes memory _message) internal view returns (G1Point memory) {
        uint256 h = uint256(keccak256(_message));
        return curveMul(P1(), h);
    }
```
P1 is the generator point of G1. @rsinha pointed out we cannot use that implementation.

Given that the method for computing the message coordinate from a message hash has to be the same across the parings API and the possible implementations in smart contracts, this task will depend on finding a suitable algorithm for hashing that can be implemented in both places. @rsinha.

**Tasks**
- [ ] Check if the TSS Library uses the pairings-api to map points to a curve
- [ ] Decide if the logic described in this ticket should be implemented in the pairings-api or the signature-api
- [ ] Implement

Dummy Code Example:
```
Signature sign(byte[] array) {
  Hash keccak256Hash = java.library.keccak256(array);
  For (int i  = 0....i < MAX_TRIES) {
    Point point = RustLibrary.getPointOnCurve(keccak256Hash); // calls arkworks at some level
    <more rust calls - 3?>
    if (some check on point) {
         return point.toSignature();
    }
    keccak256Hash = java.library.keccak256(combine(array, i));
  }
}

### PR DESCRIPTION
**Pull request by lpetrovic05**
closes https://github.com/hashgraph/hedera-services/issues/15451

### CODE REVIEW

File: cryptography/hedera-cryptography-altbn128/src/main/rust/bn254_group_elements.rs [hunk 1]
```diff
@@ -54,16 +56,48 @@ pub extern "system" fn Java_com_hedera_cryptography_altbn128_adapter_jni_ArkBn25
...
+/// * `input_x`   a 256-bit byte array that represents the x coordinate of the group element
+/// * `group_id`  in which group to perform the operation
+/// * `output`    the byte array that will be filled with the resulting group element, the size
+///               depends on the group. it will be unchanged in case the point is not in the curve
+/// # Returns
+/// *   0    Success
+/// *   -4   Business Error: Point is not in the curve
+/// * A less than 0 error code in case of error
+#[no_mangle]
+pub extern "system" fn Java_com_hedera_cryptography_altbn128_adapter_jni_ArkBn254Adapter_groupElementsFromHash(
```
mxtartaglia-sl: The name should match the native interface:
`Java_com_hedera_cryptography_altbn128_adapter_jni_ArkBn254Adapter_groupElementsFromXCoordinate`

cryptography/hedera-cryptography-altbn128/src/test/java/com/hedera/cryptography/altbn128/AltBn128GroupTest.java
```diff
@@ -78,13 +78,13 @@ void createGroupElementFromRandomIsNotNull() {
...
     @Test
-    void createGroupElementFromHashIsNotNull() {
+    void createGroupElementHashToCurveIsNotNull() {
         var group = new AltBn128Group(AltBN128CurveGroup.GROUP2);
         Random rng = new SecureRandom();
         final byte[] message = new byte[1024];
         rng.nextBytes(message);
 
-        assertNotNull(group.fromHash(message));
+        assertNotNull(group.hashToCurve(message));
```
mxtartaglia-sl, line 87: The chance that this test becomes flaky is really low, but as we are using a random seed, we might encounter a situation where we don't get a point, so the assertion doesn't stand for a specific run. 
Either we only use points we know for sure won't return null, or we are optimistic and consider that we won't find such a case ( but still, I would print the value that discovered the unusable hash for future references).

cryptography/hedera-cryptography-altbn128/src/main/rust/jni_helpers.rs
```diff
@@ -113,13 +110,13 @@ pub fn extract_random_seed(env: &JNIEnv, input_seed: &JByteArray) -> Result<[u8;
     Ok(seed_array)
 }
 
-/// Utility function to write the serialized representation of a point in an existing JByteArray
-pub fn write_return_point<G: CurveGroup>(
+/// Utility function to write the serialized representation in an existing JByteArray
+pub fn serialize_to_jbytearray<G: CanonicalSerialize>(
```
mxtartaglia-sl, line 114: nice!

cryptography/hedera-cryptography-altbn128/src/main/rust/group_element_utils.rs
```diff
@@ -99,3 +105,44 @@ pub fn group_elements_deserialize_and_validate<G: CurveGroup>(
...
+    if G::BaseField::extension_degree() == 1 {
+        // if the order is 1 the op is simpler
+        let f = <G::BaseField as Field>::BasePrimeField::from_be_bytes_mod_order(
+            &candidate_hash,
+        );
+        G::BaseField::from_base_prime_field_elems(&[f])
+    } else {
+        G::BaseField::from_random_bytes(&candidate_hash)
+    }
+}
```
mxtartaglia-sl: This method is more generic than its name suggests. It takes a candidate byte array and returns a BaseField representation. The base field element can then be used as an x or y coordinate, but that's up to the caller.

```suggestion
/// Converts a u8 array into the BaseField representation of the group (the field where the coordinates are represented) if there is such representation
pub fn coordinate_from_hash<G: SWCurveConfig>( candidate: &[u8])->Option<G::BaseField>{
    if G::BaseField::extension_degree() == 1 {
        // if the order is 1 the op is simpler
        let f = <G::BaseField as Field>::BasePrimeField::from_be_bytes_mod_order(
            &candidate,
        );
        G::BaseField::from_base_prime_field_elems(&[f])
    } else {
        G::BaseField::from_random_bytes(&candidate)
    }
}
```

mxtartaglia-sl (commented):
Some nits comments

cryptography/hedera-cryptography-altbn128/src/main/java/com/hedera/cryptography/altbn128/adapter/jni/ArkBn254Adapter.java
```diff
@@ -175,6 +175,8 @@ public static ArkBn254Adapter getInstance() {
      */
     public native int groupElementsFromSeed(final int group, final byte[] input, final byte[] output);
 
+    public native int groupElementsFromXCoordinate(final int group, final byte[] input, final byte[] output);
```
mxtartaglia-sl, line 189: missing javadoc

cryptography/hedera-cryptography-altbn128/src/main/java/com/hedera/cryptography/altbn128/AltBn128Group.java
```diff
@@ -89,8 +104,26 @@ public GroupElement random(@NonNull final byte[] seed) {
...
+        byte[] candidate = input;
+        for (int i = 0; i < HASH_RETRIES; i++) {
+            digest.update(candidate);
+            candidate = digest.digest();
+            final byte[] element = facade.fromXCoordinate(candidate);
+            if (element != null) {
+                return new AltBn128GroupElement(this, element);
+            }
+        }
+        throw new RuntimeException("Could not find a valid group element after %d tries".formatted(HASH_RETRIES));
```
mxtartaglia-sl: Can we throw an AltBn128Exception here to reduce the scope of the exceptions clients must handle?

@rsinha, how often can we encounter a situation where we cannot hash a value with this algorithm?

That scenario would fail in the consensus node, and we should know what to do.

cryptography/hedera-cryptography-altbn128/src/main/rust/bn254_group_elements.rs
lpetrovic05: yeah, missed that in the rename

cryptography/hedera-cryptography-altbn128/src/test/java/com/hedera/cryptography/altbn128/AltBn128GroupTest.java
lpetrovic05: makes sense, I hardcoded the x coordinate

cryptography/hedera-cryptography-altbn128/src/main/java/com/hedera/cryptography/altbn128/adapter/jni/ArkBn254Adapter.java
lpetrovic05: I copied the javadoc from the interface because of consistency, but I think these javadocs should not be duplicated. we could use `@inheritDoc`, but I think even that is not necessary

cryptography/hedera-cryptography-altbn128/src/main/java/com/hedera/cryptography/altbn128/AltBn128Group.java
lpetrovic05: changed

mxtartaglia-sl (approved):

rbarker-dev (approved):
Review applies to `hedera-dependency-versions/build.gradle.kts`

LGTM

PR #116, opened 2024-11-05
ci: Update setup-zig to set use-cache to false

### LINKED ISSUE #115: disable caching on setup-zig action
Running into issues with setup-zig:

```plaintext
Failed to save: Unable to reserve cache with key setup-zig-tarball-zig-linux-x86_64-0.13.0, another job may be creating this cache. More details: Cache already exists. Scope: refs/pull/111/merge, Key: setup-zig-tarball-zig-linux-x86_64-0.13.0, Version: f8c830ea35e811240be9d5006cf70b8d772b24706f3caffa189d7df216131715
fetch took 186402 m
```

Per the setup-zig [documentation](https://github.com/mlugg/setup-zig#:~:text=If%20necessary%2C%20the%20caching%20of%20the%20global%20Zig%20cache%20directory%20can%20be%20disabled%20by%20setting%20the%20option%20use%2Dcache%3A%20false.%20Don%27t%20do%20this%20without%20reason%3A%20preserving%20the%20Zig%20cache%20will%20typically%20speed%20things%20up%20and%20decrease%20the%20load%20on%20GitHub%27s%20runners.) we should be able to disable the cache:

```plaintext
If necessary, the caching of the global Zig cache directory can be disabled by setting the option use-cache: false. Don't do this without reason: preserving the Zig cache will typically speed things up and decrease the load on GitHub's runners.
```

Attempt to disable caching on setup-zig to decrease likelihood of error on download.

rbarker-dev:
**Description**:

Set the use-cache option to false to prevent cache conflicts

**Related Issue(s)**:

Fixes #115

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -58,6 +58,8 @@ jobs:
 
       - name: Setup Zig (Linux/Mac cross-compile)
         uses: mlugg/setup-zig@a67e68dc5c8281d9608136d3d7ca1b282213e4ac  # v1.2.1
+        with:
+          use-cache: 'false'
 
       - name: Setup lld (Windows cross-compile)
         run: |


andrewb1269 (approved):

rbarker-dev:
Set up zig passed the first round; planning to rebuild this three times before I call it tested.

PR #121, opened 2024-11-08
feat: 16513 ShareExtractor & ParticipantPrivateInfo

mxtartaglia-sl:
**Description**:
This PR aims to solve necessary preconditions for the extraction of shares after discussing API changes with @lpetrovic05 
The main changes are the following: 
 - The `participantDirectory` no longer holds the participant's private information. That is extracted to a different object only sent in the operation that requires the information.
 - The extraction of shares happens in a stateful object that can keep track of the process. Users might be able to configure an async version of the processing of the shares (TODO)

**Related issue(s)**:

Fixes https://github.com/hashgraph/hedera-services/issues/16513

**Notes for reviewer**:
Depending on this task:
*  https://github.com/hashgraph/hedera-services/issues/15499
* https://github.com/hashgraph/hedera-services/issues/15500

**Checklist**

- [ ] Documented (Code comments, README, etc.)
- [ ] Tested (unit, integration, etc.)

PR #122, opened 2024-11-09
feat: 16481 support participantId as long

### LINKED ISSUE #16481: We need to support `Long` value as participant ids while keep suporting Integer.MAX_SIZE of shares*participants

mxtartaglia-sl:
**Description**:
This pull request allows us to completely support Long values as participant ids, while keeping the storage of the information in an array.
**Related issue(s)**: 
this PR depends on: https://github.com/hashgraph/hedera-cryptography/pull/121

Fixes https://github.com/hashgraph/hedera-services/issues/16481

### DIFF
--- a/cryptography/hedera-cryptography-tss/src/main/java/com/hedera/cryptography/tss/api/TssParticipantDirectory.java
+++ b/cryptography/hedera-cryptography-tss/src/main/java/com/hedera/cryptography/tss/api/TssParticipantDirectory.java
@@ -19,7 +19,7 @@ package com.hedera.cryptography.tss.api;
 import static java.util.Objects.requireNonNull;
 
 import com.hedera.cryptography.bls.BlsPublicKey;
-import com.hedera.cryptography.tss.extensions.TssEncryptionKeyMap;
+import com.hedera.cryptography.tss.extensions.TssParticipantAssigmentMap;
 import edu.umd.cs.findbugs.annotations.NonNull;
 import java.util.Arrays;
 import java.util.HashMap;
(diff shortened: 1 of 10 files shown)


edward-swirldslabs (dismissed):

cryptography/hedera-cryptography-tss/src/main/java/com/hedera/cryptography/tss/api/TssParticipantDirectory.java
```diff
@@ -218,31 +195,40 @@ public TssParticipantDirectory build() {
...
+                tssEncryptionPublicKeyTable[i] = entry.tssEncryptionPublicKey;
+
+                // ParticipantId-->ParticipantIndex
+                participantIndexes.put(participantIds[i], i);
+                // ShareIndex-->ParticipantIndex
+                Arrays.fill(shareOwnersTable, currentIndex, currentIndex + entry.shareCount(), i);
+                // ParticipantIndex-->First Share; Share count
+                participantShares[i][0] = currentIndex + 1; // adds the first share and the last share
+                participantShares[i][1] = entry.shareCount();
+
```
lpetrovic05: it would be better (if possible) for any data structures related to `TssParticipantAssigmentMapping` to be built in its constructor

mxtartaglia-sl: Good catch. Yes, there is no need to do it externally.

lpetrovic05 (approved):

PR #132, opened 2024-11-14
feat: serialization tests

### LINKED ISSUE #16377: Investigate the effects of Arkworks flags in groupElements: https://github.com/arkworks-rs/algebra/issues/708

lpetrovic05:
closes https://github.com/hashgraph/hedera-services/issues/16377


cryptography/hedera-cryptography-altbn128/src/main/java/com/hedera/cryptography/altbn128/facade/FieldFacade.java
```diff
@@ -100,6 +100,10 @@ public byte[] fromBytes(@NonNull final byte[] representation) {
         if (Objects.requireNonNull(representation, "representation must not be null").length != size) {
             throw new IllegalArgumentException("Invalid byte[] representation");
         }
+        if ((representation[size - 1] & 0b1100000) != 0) {
+            throw new IllegalArgumentException(
+                    "Invalid byte[] representation, the number is expected to be a 254-bit number, so the 2 most significant bits must be unset");
+        }
```
mxtartaglia-sl: Still not on board with this check

mxtartaglia-sl: Are the bits really unused? What did the test show? What happens if you compare two FieldElement values, one with and the other without each of those flipped bytes? 

I say this because this change broke tests in AltBn128FieldElementTest.
Now I am more than convinced that we should not do this.

lpetrovic05: I have removed this constraint. I think this functionality should exist, but it should be separate from deserialization. We should discuss tomorrow.

rbarker-dev (approved):

cryptography/hedera-cryptography-altbn128/src/main/java/com/hedera/cryptography/altbn128/facade/FieldFacade.java
mxtartaglia-sl: The only assurance we need is that if, for example:
`a = 10000000000...00001011`  (binary representation of a fieldElement with the bits that allegedly are used for flags set to 1), and `b = 00000000000...00001000` (same as a but with the bits that allegedly are used for flags set to 0), then `a != b`. If that check is true, the rest is an unnecessary complication and incomplete implementation. 

What I mean by the last statement is that in the complete scalar field, even when using 254-bit numbers, we find instances where `fromBytes(someByteArray)` produces an element where `element.representation != someByteArray` is true.
That shows that even if we think this is necessary, counting the number of bits in the input is not a complete implementation to detect an input that would produce an "_incorrect_" fieldElement. On top of that, we would have to execute this check every time we create a fieldElement,  which makes it a complication.

mxtartaglia-sl (approved):

cryptography/hedera-cryptography-altbn128/src/main/java/com/hedera/cryptography/altbn128/facade/FieldFacade.java
lpetrovic05: I agree, this check has been removed

PR #156, opened 2024-11-27
!feat(16681): API for Serialization

### LINKED ISSUE #16681: Define an API for serialization
Come up with a API for serializing:
- Elements
- Keys
- Signatures
- Tss messages

mxtartaglia-sl:
**Description**:
The serialization mechanism for the keys is being reworked based on a decision to be compatible with Ethereum's format. 
This PR introduces changes to the serialization/deserialization mechanisms to allow external users to implement their serialization/deserialization mechanisms, and it also provides some default Serializer/Deserializer mechanisms to provide out-of-the-box conversion from/to byte arrays.

⚠️  Breaking changes: 
* `com.hedera.cryptography.tss.extensions` is now a public module package. All classes that belong to the protocol implementation were moved to `com.hedera.cryptography.tss.impl`, now a private module package. Consequently, `Groth21Service` is hidden from external users, and the way to construct an instance of the service is by calling: `TssService.createDefaultService(schema, random)`.

🚫  Deprecation Note:
*  The previous format for serializing `BlsPrivateKey`, `BlsPublicKey`s,  `BlsSignature`s is deprecated. The API still supports the previous values in the files, but that support will be removed. The new format compatible with ethereum is still not included in this version.

*   `BlsPrivateKey`, `BlsPublicKey`s,  `BlsSignature`s and `TssMessage`s  `toBytes` and `fromBytes` methods are now  `@Deprecated`. Users should move to instances of serializers/deserializers by implementing a custom mechanism or using the defaults accordingly. FYI:
** `com.hedera.cryptography.tss.extensions.serialization.DefaultTssMessageSerialization`
** `com.hedera.cryptography.bls.extensions.serialization.DefaultBlsPrivateKeySerialization`
** `com.hedera.cryptography.bls.extensions.serialization.DefaultBlsPublicKeySerialization`
** `com.hedera.cryptography.bls.extensions.serialization.DefaultBlsSignatureSerialization`

**Related issue(s)**:

Fixes: https://github.com/hashgraph/hedera-services/issues/16681


cryptography/hedera-cryptography-altbn128/src/main/java/com/hedera/cryptography/altbn128/facade/GroupFacade.java
```diff
@@ -204,7 +202,9 @@ public byte[] fromBytes(@NonNull final byte[] bytes) {
             throw new AltBn128Exception(result, "groupElementsBytes in" + this.group);
         }
 
-        return bytes;
+        byte[] ret = new byte[bytes.length];
```
lpetrovic05: final

cryptography/hedera-cryptography-bls/src/main/java/com/hedera/cryptography/asciiarmored/AsciiArmoredFiles.java
```diff
@@ -46,16 +47,19 @@ private AsciiArmoredFiles() {
...
      *
      * @param path The location of the file in the fileSystem
+     * @param deserializer the deserializer to use
      * @return the private key contained in the file
      * @throws IOException In case of file reading error
      */
     @NonNull
-    public static BlsPrivateKey readPrivateKey(@NonNull final Path path) throws IOException {
+    public static BlsPrivateKey readPrivateKey(
+            @NonNull final Path path, @NonNull final Deserializer<BlsPrivateKey> deserializer) throws IOException {
```
lpetrovic05, line 77: Maybe we overload this method with a default deserializer, since I know services is already using this method

cryptography/hedera-cryptography-bls/src/main/java/com/hedera/cryptography/asciiarmored/AsciiArmoredFiles.java
```diff
@@ -88,34 +92,24 @@ private static AsciiArmoredFile asciiArmoredRead(@NonNull final Path path) throw
...
+     * @param <T> target type
      * @throws IOException In case of file writing error
      */
-    public static void writeKey(@NonNull final Path path, @NonNull final BlsPublicKey publicKey) throws IOException {
-        writeKey(
+    public static <T> void write(
+            @NonNull final Path path,
+            final Serializer<T> serializer,
+            @NonNull final T key,
+            @NonNull final AsciiArmoredType type)
```
lpetrovic05: why introduce this new `type` variable?

cryptography/hedera-cryptography-bls/src/main/java/com/hedera/cryptography/bls/extensions/serialization/DefaultBlsPrivateKeySerialization.java
```diff
@@ -0,0 +1,112 @@
...
+     */
+    private static class DefaultSerializer implements Serializer<BlsPrivateKey> {
+        private final Serializer<FieldElement> elementSerializer;
+
+        /**
+         * Constructor.
+         */
+        DefaultSerializer() {
+            this.elementSerializer = DefaultFieldElementSerialization.getSerializer();
+        }
```
lpetrovic05: making it a record will make it less verbose

cryptography/hedera-cryptography-bls/src/main/java/com/hedera/cryptography/asciiarmored/AsciiArmoredFiles.java
mxtartaglia-sl: Because the key is now generic, and you can reuse it for both private and public use, you need to specify what you are writing.
I can do better here, maybe all the 3 parameters can be condensed in a enum

cryptography/hedera-cryptography-bls/src/main/java/com/hedera/cryptography/bls/extensions/serialization/DefaultBlsPrivateKeySerialization.java
mxtartaglia-sl: yes, agreed, removed the record apoproach after our conversation. but will put it back

cryptography/hedera-cryptography-tss/src/main/java/com/hedera/cryptography/tss/extensions/Groth21Service.java
```diff
@@ -74,16 +75,31 @@ public TssServiceRekeyStage rekeyStage() {
...
+    @NonNull
+    @Override
+    public <S> TssMessage messageFrom(
+            @NonNull final Transformer<S, TssMessage> tssMessageTransformer, @NonNull final S message)
+            throws TssMessageParsingException {
         try {
-            return Groth21Message.fromBytes(message, tssParticipantDirectory, signatureSchema);
+            return tssMessageTransformer.transform(message);
         } catch (Exception e) {
-            throw new TssMessageParsingException("Could not read tss message", e);
```
lpetrovic05: why even have this method? whoever calls this method can do the transformation on their own

cryptography/hedera-cryptography-bls/src/main/java/com/hedera/cryptography/asciiarmored/AsciiArmoredFiles.java
mxtartaglia-sl: True, but here is the thing: even if we are compatible with the API, the keys we write to the file and how we read them will not be compatible.

cryptography/hedera-cryptography-tss/src/main/java/com/hedera/cryptography/tss/extensions/Groth21Service.java
mxtartaglia-sl: For documentation purposes only? I don't know; it feels like the whole serializer/transformer approach is disconnected from the solution, and without previous knowledge that you need to do that, nothing in the API forces/or suggests you use them.

cryptography/hedera-cryptography-bls/src/main/java/com/hedera/cryptography/asciiarmored/AsciiArmoredFiles.java
mxtartaglia-sl: fixed it by overloading the method

cryptography/hedera-cryptography-bls/src/main/java/com/hedera/cryptography/asciiarmored/AsciiArmoredFiles.java
lpetrovic05: we could implement the arkworks serialization and make this work, since there are PRs already using it, i think we should

lpetrovic05 (approved):

andrewb1269 (approved):
Review applies to `version.txt` only. Please ensure RE reviews this PR as well, if appropriate.

PR #183, opened 2024-12-19
ci: disable cross-version tests in Dependabot PRs

jjohannes:
These require the 'Build software' step to write to the Gradle remote cache. But Dependabot PRs to not have write access to the cache.

I think this is a reasonable compromise as, if an updates breaks something, it is most likely in Java code.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -72,6 +72,7 @@ jobs:
 
   test-cross-platform:
     name: Test on ${{ matrix.os }}
+    if: ${{ github.actor != 'dependabot[bot]' }}
     needs: build
     strategy:
       matrix:


nathanklick (changes requested):
Why not just add the Gradle Cache credentials to the dependabot secrets?

jjohannes:
> Why not just add the Gradle Cache credentials to the dependabot secrets?

If that is possible, that's even besser. Can you do that @andrewb1269hg and then we can discard this PR?

andrewb1269:
> > Why not just add the Gradle Cache credentials to the dependabot secrets?
> 
> If that is possible, that's even besser. Can you do that @andrewb1269hg and then we can discard this PR?

Yes, I believe that's possible. I'll write an issue to capture this effort and we can close this PR out by tying it to the new issue.

andrewb1269:
@jjohannes I have added `GRADLE_CACHE_USERNAME` and `GRADLE_CACHE_PASSWORD` to the dependabot secrets. These should be available for usage.

jjohannes:
Thanks @andrewb1269hg. Credentials work as this PR shows https://github.com/hashgraph/hedera-cryptography/pull/175

Closing this one.

PR #206, opened 2025-01-14
ci: update semantic release workflows

### LINKED ISSUE #193: Implement semantic release
We will need to implement semantic release within the hedera-cryptography repo to facilitate versioning.

PavelSBorisov:
**Description**:
This PR implements semantic release within the hedera-cryptography repo to facilitate versioning.

**Related issue(s)**:

Fixes #193 

**Notes for reviewer**:
N/A

**Checklist**


PavelSBorisov:
@andrewb1269hg @rbarkerSL, two of the secrets used in the workflow files currently don't exist:
```
${{ secrets.GIT_USER_NAME }}
${{ secrets.GIT_USER_EMAIL }}
```
You will need to create them before this can be tested and merged.

nathanklick (dismissed):
We do not use and should not use the build identifier model 9n this repository. That is only needed for CITR on the `hedera-services` repo.

.github/workflows/zxf-version-roll.yaml
```diff
@@ -0,0 +1,107 @@
...
+          echo "::group::Download SemVer Binary"
+          sudo curl -L -o /usr/local/bin/semver https://raw.githubusercontent.com/fsaintjacques/semver-tool/master/src/semver
+          echo "::endgroup::"
+          echo "::group::Change SemVer Binary Permissions"
+          sudo chmod -v +x /usr/local/bin/semver
+          echo "::endgroup::"
+          echo "::group::Show SemVer Binary Version Info"
+          semver --version
+          echo "::endgroup::"
+      - name: Read Current Version
```
andrewb1269:
```suggestion

      - name: Read Current Version
```

.github/workflows/zxf-version-roll.yaml
```diff
@@ -0,0 +1,107 @@
...
+        id: read_version
+        run: |
+          current_version=$(cat version.txt)
+          echo "Current version: $current_version"
+          echo "version=$current_version" >> $GITHUB_ENV
+          printf "Current Version: $current_version" >> $GITHUB_STEP_SUMMARY
+      - name: Increment Minor Version
```
andrewb1269:
```suggestion

      - name: Increment Minor Version
```

.github/workflows/zxf-version-roll.yaml
```diff
@@ -0,0 +1,107 @@
...
+          # Strip the -SNAPSHOT suffix
+          base_version=$(echo $version | sed 's/-SNAPSHOT//')
+          # Increment the minor version
+          new_version=$(semver bump minor $base_version)
+          # Add the -SNAPSHOT suffix back
+          new_version="${new_version}-SNAPSHOT"
+          echo "New version: $new_version"
+          echo "new_version=$new_version" >> $GITHUB_ENV
+          printf "New Version: new_version" >> $GITHUB_STEP_SUMMARY
+      - name: Update version.txt
```
andrewb1269:
```suggestion

      - name: Update version.txt
```

.github/workflows/zxf-version-roll.yaml
```diff
@@ -0,0 +1,107 @@
...
+        run: |
+          echo $new_version > version.txt
+      - name: Publish to Maven Central
```
andrewb1269:
```suggestion

      - name: Publish to Maven Central
```

.github/workflows/zxf-version-roll.yaml
```diff
@@ -0,0 +1,107 @@
...
+        run: |
+          ./gradlew versionAsSpecified -PnewVersion=$new_version
+      - name: Add & Commit
```
andrewb1269:
```suggestion

      - name: Add & Commit
```

.github/workflows/zxf-trigger-semantic-release.yaml
```diff
@@ -0,0 +1,107 @@
...
+      - name: Setup Node
+        uses: actions/setup-node@39370e3970a6d050c480ffad4ff0ed4d3fdee5af # v4.1.0
+        with:
+          node-version: 20
+
+      - name: Install Dependencies
+        run: |
+          echo "Installing semantic-release"
+          npm install -g semantic-release
+      - name: Configure Git for Semantic Release
```
andrewb1269:
```suggestion

      - name: Configure Git for Semantic Release
```

.github/workflows/zxf-trigger-semantic-release.yaml
```diff
@@ -0,0 +1,107 @@
...
+        run: |
+          git config user.name "${{ secrets.GIT_USER_NAME }}"
+          git config user.email "${{ secrets.GIT_USER_EMAIL }}"
+      - name: Run Semantic Release
```
andrewb1269:
```suggestion

      - name: Run Semantic Release
```

andrewb1269:
> @andrewb1269hg @rbarkerSL, two of the secrets used in the workflow files currently don't exist:
> 
> ```
> ${{ secrets.GIT_USER_NAME }}
> ${{ secrets.GIT_USER_EMAIL }}
> ```
> 
> You will need to create them before this can be tested and merged.

I added these two requested secrets to the repository.

PavelSBorisov:
> We do not use and should not use the build identifier model 9n this repository. That is only needed for CITR on the `hedera-services` repo.

@nathanklick You're right, just saw that we don't do any builds here. 
What should the process look like in this repo, should we be passing a commit SHA in place of the build id?

PavelSBorisov:
After a discussion with @andrewb1269hg yesterday, we settled on automatically sourcing the new semantic release version from the latest commit on main.
No inputs should be passed to the trigger release workflow.

andrewb1269:
@PavelSBorisov: @rbarkerSL and I have a meeting with a few folks next Wednesday to confirm this is the path they'd like to go down for publishing semantic release versions of the package.

Let's hold off on reviewing and merging this code for now.

.github/workflows/publish-official-release.yaml
```diff
@@ -0,0 +1,152 @@
...
+        with:
+          author_name: ${{ secrets.GIT_USER_NAME }}
+          author_email: ${{ secrets.GIT_USER_EMAIL }}
+          message: "chore: update version.txt"
+          add: "version.txt"
+          commit: --signoff
+
+      # This step is used to update the version number in the build.properties file.
+      - name: Apply Gradle Version Number Update (Explicit)
+        run: ./gradlew versionAsSpecified -PnewVersion="$NEW_VERSION"
```
PavelSBorisov: @andrewb1269hg another thing here, could you please confirm if this step is actually needed since to my understanding this gradle task supposedly updates the version.txt file which is already done above? (I think the comment is outdated)

.github/workflows/publish-snapshot-release.yml
```diff
@@ -0,0 +1,108 @@
...
+          java-version: ${{ env.JAVA_VERSION }}
+
+      - name: Setup Gradle
+        uses: gradle/actions/setup-gradle@0bdd871935719febd78681f197cd39af5b6e16a6 # v4.2.2
+        with:
+          gradle-version: ${{ env.GRADLE_VERSION }}
+
+      # This step is used to update the version number to a snapshot in the version.txt file in the runner.
+      - name: Apply Version Number Update (Explicit)
+        run: ./gradlew versionAsSnapshot
```
PavelSBorisov: @andrewb1269hg does this work the way I think it does or is my implementation incorrect here?

.github/workflows/publish-official-release.yaml
andrewb1269: I confirmed we should not need this step anymore as lines 77-78 will handle this update.

.github/workflows/publish-snapshot-release.yml
andrewb1269: You can see the function definition [here in the hiero-ledger organization](https://github.com/hiero-ledger/hiero-gradle-conventions/blob/3d0ec2ec8565137250018359eb2d0b8c9401f310/src/main/kotlin/org.hiero.gradle.feature.versioning.gradle.kts#L88)

It looks like you'll need to add the  `-PnewVersion=<version>` flag to this step to update `version.txt` file.

.github/workflows/publish-official-release.yaml
```diff
@@ -0,0 +1,160 @@
...
+          passphrase: ${{ secrets.GPG_KEY_PASSPHRASE }}
+
+      - name: Setup Node
+        uses: actions/setup-node@39370e3970a6d050c480ffad4ff0ed4d3fdee5af # v4.1.0
+        with:
+          node-version: 20
+
+      - name: Install Dependencies
+        run: |
+          echo "Installing semantic-release"
```
andrewb1269, line 101: Does this step actually install semantic release? I think other repos have a larger block for installing semantic release.

For example, see [this PR](https://github.com/hashgraph/hedera-services/pull/17438/files#)
`.github/workflows/flow-trigger-release.yaml`
Line: 157
```
          npm install -g semantic-release@24.2.0 @semantic-release/git@10.0.1 @semantic-release/github@11.0.1 \
            @semantic-release/exec@6.0.3 semantic-release-helm3@2.9.3 \
            conventional-changelog-conventionalcommits@8.0.0 \
            @commitlint/cli@19.5.0 @commitlint/config-conventional@19.5.0 \
            marked-mangle@1.1.10 marked-gfm-heading-id@4.1.1 semantic-release-conventional-commits@3.0.0
```

.github/workflows/publish-official-release.yaml
```diff
@@ -0,0 +1,160 @@
...
+          npm install
+
+      - name: Configure Git for Semantic Release
+        run: |
+          git config user.name "${{ secrets.GIT_USER_NAME }}"
+          git config user.email "${{ secrets.GIT_USER_EMAIL }}"
+
+      - name: Dry Run Semantic Release
```
andrewb1269: I would propose combining the `Dry Run Semantic Release` and `Run Semantic Release` steps together into a single step and use the `dry-run-enabled` input field as a toggle for environment variables.

For example, see [this PR](https://github.com/hashgraph/hedera-services/pull/17438/files#)
`.github/workflows/flow-trigger-release.yaml`
Line: 170
```
          run: |
          ARGS=""
          if [[ "${{ github.event.inputs.dry-run-enabled }}" == true ]]; then
            ARGS="--dry-run"
          fi
          echo "Running semantic release command:"
          echo "npx semantic-release ${ARGS}
          npx semantic-release ${ARGS}
```

.github/workflows/publish-official-release.yaml
```diff
@@ -0,0 +1,160 @@
...
+          npx semantic-release --dry-run --debug
+          echo "The version number assigned to the next release is: $VERSION"
+
+      - name: Run Semantic Release
+        if: ${{ inputs.dry-run-enabled != 'true' && !cancelled() && !failure() }}
+        env:
+          GITHUB_TOKEN: ${{ secrets.GH_ACCESS_TOKEN }}
+        run: |
+          echo "Running semantic-release now"
+          npx semantic-release --debug
```
andrewb1269: I'm not sure if we should be running debug mode for mainline. Perhaps add another input field for running in debug mode, or remove the debug mode all together.

.github/workflows/publish-official-release.yaml
```diff
@@ -0,0 +1,160 @@
...
+
+      - name: Get Latest Tag and update version.txt
```
andrewb1269: The following steps can be combined into a single step:

`Get Latest Tag and update version.txt`
`Commit and Push version.txt`
`Apply Gradle Version Number Update (Explicit)`

These can be combined into a single steps where we perform the following:
1. Fetch and pull from remote to get the latest tags (please confirm `semantic release` publishes tags as well).
2. Echo to the user what the new version is based on the tags
3. Run `./gradlew versionAsSpecified -PnewVersion="$NEW_VERSION"` to update the new version number in `version.txt`
4. Add, commit, and push the new version number to remote.

We have bypass rules in place for our automation user to directly push on mainline.

.github/workflows/publish-official-release.yaml
andrewb1269: I see you've added a `package.json` file, so this comment probably no longer makes sense. Please confirm that we can use the simple `npm install` to achieve the large block I posted above.

PavelSBorisov: That's correct, I've been using npm install during local development and dry run testing so in theory this should also work in the runner environment.
I am not sure all of the packages in the block you referenced are needed in this repo, but we can add them if they prove necessary during testing.

.github/workflows/publish-official-release.yaml
PavelSBorisov: Great idea, I'll implement a version of this!

.github/workflows/publish-official-release.yaml
PavelSBorisov: Removing debug mode

.github/workflows/publish-official-release.yaml
PavelSBorisov: > Fetch and pull from remote to get the latest tags (please confirm semantic release publishes tags as well).

Confirmed semantic-release should always [publish a tag as part of its release steps](https://semantic-release.gitbook.io/semantic-release#release-steps)

> Run ./gradlew versionAsSpecified -PnewVersion="$NEW_VERSION" to update the new version number in version.txt

After doing a bit of digging in the docs it seems like this is redundant as we are utilizing the .releaserc file with:
```
"verifyRelease": [
    [
      "@semantic-release/exec",
      { "cmd": "echo ${nextRelease.version} > version.txt" }
    ]
  ],
 ```
 This outputs the new version to version.txt automatically upon running `npx semantic-release` (also does it in both dry-run and actual release mode).
 
 I'll refactor these steps.

File: .github/workflows/publish-official-release.yaml [hunk 14]
```diff
@@ -0,0 +1,183 @@
...
+        uses: EndBug/add-and-commit@a94899bca583c204427a224a7af87c02f9b325d5 # v9.1.4
+        with:
+          author_name: ${{ secrets.GIT_USER_NAME }}
+          author_email: ${{ secrets.GIT_USER_EMAIL }}
+          message: "chore: update version.txt"
+          add: "version.txt"
+          commit: --signoff
+
+      # This step is used only in dry-run mode to add -SNAPSHOT suffix to version number in version.txt
+      - name: Dry Ryn Apply Gradle Version Number Update (Explicit)
```
andrewb1269:
```suggestion
      - name: Dry Run Apply Gradle Version Number Update (Explicit)
```

andrewb1269 (changes requested):
Small typo to resolve.

andrewb1269:
@PavelSBorisov I've written up an outline of a testing procedure [here in Notion](https://www.notion.so/swirldslabs/hedera-cryptography-Semantic-Release-18e7c9ab2591808ea6c1ec1dc18316be?pvs=4). Please let me know if you have issues accessing it.

I will continue to add to the testing plan, and let's chat before you start executing it to make sure everything makes sense and is accurate.

@rbarkerSL for awareness.

andrewb1269:
@PavelSBorisov one other issue I found with this PR:
`.github/workflows/publish-snapshot-release.yml` please rename to `.github/workflows/publish-snapshot-release.yaml` to be consistent in naming (file extension).

.github/workflows/publish-snapshot-release.yml
andrewb1269: I think we want to remove this step.

When we're running the `publish-snapshot-release.yaml` workflow, we want to publish a SNAPSHOT of the existing version number in `version.txt`.

Example: Let's say we're on `v0.2.0`. If we merge a new feature (which would roll to `v0.3.0`) we do not want the first publish to be a snapshot version. We should continue to publish under `v0.2.0` until an official version of `v0.3.0` is created.

Once `v0.3.0` exists then we'll want to publish under `v0.3.0-SNAPSHOT` going forward. The only time the `version.txt` file should be modified is in the `publish-official-release.yaml` workflow.

.github/workflows/publish-snapshot-release.yml
```diff
@@ -0,0 +1,109 @@
...
+  LC_ALL: C.UTF-8
+  GRADLE_CACHE_USERNAME: ${{ secrets.GRADLE_CACHE_USERNAME }}
+  GRADLE_CACHE_PASSWORD: ${{ secrets.GRADLE_CACHE_PASSWORD }}
+  JAVA_DISTRIBUTION: temurin
+  JAVA_VERSION: 21.0.4
+  GRADLE_VERSION: wrapper
+
+jobs:
+  publish-maven-central:
+    name: Publish Maven Central
```
andrewb1269:
```suggestion
    name: Publish Snapshot Release of Hedera Cryptography
```

.github/workflows/publish-snapshot-release.yml
```diff
@@ -0,0 +1,109 @@
+# SPDX-License-Identifier: Apache-2.0
+name: "Publish Release"
```
andrewb1269:
```suggestion
name: "Publish Snapshot Release"
```

.github/workflows/publish-official-release.yaml
```diff
@@ -0,0 +1,183 @@
...
+
+env:
+  REGISTRY: ghcr.io
+  LC_ALL: C.UTF-8
+  GRADLE_CACHE_USERNAME: ${{ secrets.GRADLE_CACHE_USERNAME }}
+  GRADLE_CACHE_PASSWORD: ${{ secrets.GRADLE_CACHE_PASSWORD }}
+
+jobs:
+  publish:
+    name: Publish Semantic Release
```
andrewb1269:
```suggestion
    name: Publish Official Release of Hedera Cryptography
```

andrewb1269 (changes requested):

.github/workflows/publish-snapshot-release.yml
PavelSBorisov: In this workflow, what you are describing would be the result.
The `versionAsSnapshot` [parses the current version from the `version.txt` file and appends `-SNAPSHOT` without incrementing it](https://github.com/hiero-ledger/hiero-gradle-conventions/blob/3d0ec2ec8565137250018359eb2d0b8c9401f310/src/main/kotlin/org.hiero.gradle.feature.versioning.gradle.kts#L74) and since we aren't modifying it anywhere else in the workflow, the end result would be a snapshot of the current version.

However that wouldn't be the case in the `publish-official-release` workflow in dry-run mode. There we are using the `versionAsSnapshot` after we have already incremented the version in `version.txt` in a previous step by running `npx semantic-release` in dry-run mode, thus the result would be a snapshot of the expected new version. (as I thought we were planning to do for testing purposes)
Let me know if you think we should avoid that there.

.github/workflows/publish-official-release.yaml
```diff
@@ -0,0 +1,176 @@
...
+        type: string
+        required: false
+        default: "temurin"
+      java-version:
+        description: "Java JDK Version:"
+        type: string
+        required: false
+        default: "21.0.4"
+      branch:
+        description: "Branch to release from"
```
andrewb1269:
```suggestion
        description: "Branch to release from:"
```

.github/workflows/publish-official-release.yaml
```diff
@@ -0,0 +1,176 @@
...
+
+          if [[ "$FILE_VERSION" != "$LATEST_TAG" ]]; then
+            echo "WARNING: version.txt ($FILE_VERSION) does not match the latest tag ($LATEST_TAG)."
+            echo "Updating version.txt to match the latest tag."
+            echo "$LATEST_TAG" > version.txt
+          else
+            echo "version.txt is already up to date."
+          fi
+
+      - name: Commit and Push version.txt
```
andrewb1269:
```suggestion
      - name: Commit and Push Modified Files
```

.github/workflows/publish-snapshot-release.yaml
```diff
@@ -0,0 +1,116 @@
...
+name: "Publish Snapshot Release"
+
+on:
+  push:
+    branches:
+      - main
+  workflow_dispatch:
+    inputs:
+      branch:
+        description: "Branch to release from"
```
andrewb1269:
```suggestion
        description: "Branch to release from:"
```

.github/workflows/publish-snapshot-release.yaml
```diff
@@ -0,0 +1,116 @@
...
+          egress-policy: audit
+
+      - name: Checkout Code
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+        with:
+          fetch-depth: 0
+          ref: ${{ inputs.branch }}
+
+      # This step is required for some self-hosted runners because the default gpg does not include the gpg2 command.
+      - name: Install GnuPG Tools
```
andrewb1269, line 49: Do we need to include this step in the `publish-official-release` workflow as well?

.releaserc
```diff
@@ -1,63 +1,62 @@
...
+  ],
   "plugins": [
-    "@semantic-release/commit-analyzer",
     [
-      "@semantic-release/release-notes-generator",
+      "@semantic-release/commit-analyzer",
       {
-        "preset": "conventionalcommits"
+        "preset": "conventionalcommits",
+        "presetConfig": {
```
andrewb1269: Can we change the `presetConfig` to be something default rather than listing out the section below?

.releaserc
```diff
@@ -1,63 +1,62 @@
...
     [
-      "@semantic-release/exec",
-      { "cmd": "echo ${nextRelease.version} > VERSION" }
-    ]
-  ],
-  "prepare": [
+      "@semantic-release/release-notes-generator",
+      {
+        "preset": "conventionalcommits",
+        "presetConfig": {
```
andrewb1269: Can we change the `presetConfig` to be something default rather than listing out the section below?

andrewb1269 (changes requested):

.github/workflows/publish-official-release.yaml
```diff
@@ -0,0 +1,176 @@
...
+
+defaults:
+  run:
+    shell: bash
+
+permissions:
+  id-token: write
+  contents: write
+  actions: write
+  pages: write
```
rbarker-dev: Why do we need pages write?

.github/workflows/publish-snapshot-release.yaml
```diff
@@ -0,0 +1,116 @@
...
+      # if one or more components fails to build then it could result in a partial publish to Maven Central.
+      # The inclusion of this step ensures the job fails cleanly without publishing to Maven Central if any of the
+      # components fail to build.
+      - name: Gradle Assemble
+        id: gradle-build
+        if: ${{ !cancelled() && !failure() }}
+        run: ./gradlew assemble
+
+      # Publishes the artifacts to the Maven Central Nexus staging repository.
+      # TODO for initial release and testing, we use -PpublishTestRelease to not automatically release from staging.
```
rbarker-dev: this goes away before merge?

.releaserc
rbarker-dev: These are the conventionalCommit tags that are picked up. This set is a standard set from conventional commits and it includes the tags we want to see.

.releaserc
rbarker-dev: These are the conventionalCommit tags that are picked up. This set is a standard set from conventional commits and it includes the tags we want to see.

.github/workflows/publish-official-release.yaml
PavelSBorisov: Possibly leftover permissions from the troubleshooting process, I'll clean them up

.github/workflows/publish-snapshot-release.yaml
PavelSBorisov: Yes, I'll add it for uniformity

.releaserc
PavelSBorisov: It should indeed be the standard set but I only added it because of the plugins' docs, which stated that a presetConfig was necessary for conventionalcommits - This turns out to be false, after testing on my private repo, the plugin does load the defaults if no presetconfig is specified which means this can be removed.

.releaserc
PavelSBorisov: It should indeed be the standard set but I only added it because of the plugins' docs, which stated that a presetConfig was necessary for conventionalcommits - This turns out to be false, after testing on my private repo, the plugin does load the defaults if no presetconfig is specified which means this can be removed.

.github/workflows/publish-snapshot-release.yaml
PavelSBorisov: yep, removed the -PpublishTestRelease flag from the command and comment line

andrewb1269:
@PavelSBorisov there are 2 other things we need to do before merging:

1. Ensure we have applied a "previous" tag of v0.2.0 to the mainline. This will be our starting point, then we can cut an official release with the appropriate automatic version roll forward.

2. Ensure various override permissions in the rules are cleaned up correctly.

Also @rbarker-dev will be performing the review on this since you and I worked so closely on implementing it.

PavelSBorisov:
> 1. Ensure we have applied a "previous" tag of v0.2.0 to the mainline. This will be our starting point, then we can cut an official release with the appropriate automatic version roll forward.

@andrewb1269hg What about the existing tags and releases in the repo, shouldn't our "starting point" be the last published release tag on main (v1.7.0) at the moment?

.github/workflows/publish-snapshot-release.yaml
```diff
@@ -82,7 +81,7 @@ jobs:
           gradle-version: ${{ env.GRADLE_VERSION }}
 
       # This step is used to update the version number to a snapshot in the version.txt file in the runner.
-      - name: Apply Version Number Update (Explicit)
+      - name: Add snapshot suffix to current version
```
andrewb1269:
```suggestion
      - name: Add SNAPSHOT Suffix to Current Version
```

andrewb1269 (changes requested):

.github/workflows/publish-official-release.yaml
```diff
@@ -0,0 +1,185 @@
...
+          # as semantic-release doesn't support headless annotated signed tags and
+          # the process hangs indefinitely awaiting input in the dialogue
+          EDITOR: "echo 'Created-by: semantic-release' > $1"
+        run: |
+          if [[ "${{ inputs.dry-run-enabled }}" == true ]]; then
+            echo "Dry Running semantic-release now..."
+            npx semantic-release --dry-run
+          else
+            echo "Running semantic-release now..."
+            npx semantic-release --debug
```
andrewb1269:
```suggestion
            npx semantic-release
```

andrewb1269 (dismissed):

.github/workflows/publish-snapshot-release.yaml
```diff
@@ -0,0 +1,114 @@
...
+      # Technically, this step is not required because the release steps will build what is required; however,
+      # if one or more components fails to build then it could result in a partial publish to Maven Central.
+      # The inclusion of this step ensures the job fails cleanly without publishing to Maven Central if any of the
+      # components fail to build.
+      - name: Gradle Assemble
+        id: gradle-build
+        if: ${{ !cancelled() && !failure() }}
+        run: ./gradlew assemble
+
+      # Publishes the artifacts to the Maven Central Nexus staging repository.
```
jjohannes:
```suggestion
      # Publishes the artifacts to the Maven Central Nexus snapshot repository.
```

andrewb1269:
This looks good to me, but I need @rbarker-dev to do the review. Once he reviews I'll remove my block (I don't want to approve and make this able to merge) then @PavelSBorisov can merge it to the mainline.

.github/workflows/publish-official-release.yaml
```diff
@@ -0,0 +1,185 @@
...
+      java-version:
+        description: "Java JDK Version:"
+        type: string
+        required: false
+        default: "21.0.4"
+      branch:
+        description: "Branch to release from:"
+        type: string
+        required: false
+        default: "main"
```
rbarker-dev:
```suggestion
      ref:
        description: "Github ref (branch or named tag) to release from:"
        type: string
        required: false
        default: "main"
```

.github/workflows/publish-official-release.yaml
```diff
@@ -0,0 +1,185 @@
...
+        uses: step-security/harden-runner@c95a14d0e5bab51a9f56296a4eb0e416910cd350 # v2.10.3
+        with:
+          egress-policy: audit
+
+      - name: Checkout Code
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+        with:
+          token: ${{ secrets.GH_ACCESS_TOKEN }}
+          fetch-depth: 0
+          ref: ${{ inputs.branch }}
```
rbarker-dev, line 57:
```suggestion
          ref: ${{ inputs.ref }}
```

.github/workflows/publish-official-release.yaml
```diff
@@ -0,0 +1,185 @@
...
+          GIT_AUTHOR_NAME: ${{ steps.gpg.outputs.name}}
+          GIT_AUTHOR_EMAIL: ${{ steps.gpg.outputs.email}}
+          GIT_COMMITTER_NAME: ${{ steps.gpg.outputs.name}}
+          GIT_COMMITTER_EMAIL: ${{ steps.gpg.outputs.email}}
+          # The EDITOR var is required to ensure the tag is signed with the GPG key
+          # as semantic-release doesn't support headless annotated signed tags and
+          # the process hangs indefinitely awaiting input in the dialogue
+          EDITOR: "echo 'Created-by: semantic-release' > $1"
+        run: |
+          if [[ "${{ inputs.dry-run-enabled }}" == true ]]; then
```
rbarker-dev: does this pass? It looks like it's a string compared to a boolean value. I think i would expect it to be `[[ "${{ inputs.dry-run-enabled }}" == "true" ]]`

No change required just verifying this part has been tested.

rbarker-dev (changes requested):

.github/workflows/publish-official-release.yaml
PavelSBorisov: Yes, tested in latest dry-run test https://github.com/hashgraph/hedera-cryptography/actions/runs/13306659441/job/37159090949

rbarker-dev:
```suggestion
          if [[ "${{ inputs.dry-run-enabled }}" == "true" ]]; then
```

Okay, I think we should do this.

andrewb1269:
@PavelSBorisov let me know once other folks from the Platform CI groups have approved this. I'll remove my block once others approve 👍

.github/workflows/publish-official-release.yaml
PavelSBorisov: Okay, I'll add this commit and do a test dry run

PavelSBorisov: Tested and it looks good - https://github.com/hashgraph/hedera-cryptography/actions/runs/13671461633/job/38222413717

rbarker-dev (dismissed):

rbarker-dev:
Make sure to update to remove merge conflicts but then good to go @PavelSBorisov @andrewb1269hg

rbarker-dev (approved):

andrewb1269 (dismissed):

andrewb1269 (changes requested):
Rework to discuss on March 11 meeting.

PavelSBorisov:
superseded by https://github.com/hashgraph/hedera-cryptography/pull/303

PR #227, opened 2025-01-16
feat: introduce HintsLibraryBridge

anthony-swirldslabs:
**Description**:
Introducing the `HintsLibraryBridge` class that implements a JNI bridge for the HintsLibrary implementation. Please note that the corresponding native Rust methods will be implemented in a separate PR under a separate story: https://github.com/hashgraph/hedera-cryptography/issues/218

**Related issue(s)**:

Fixes #217 

**Notes for reviewer**:
`gr build` succeeds.

**Checklist**

- [x] Documented (Code comments, README, etc.)
- [ ] Tested (unit, integration, etc.)

PR #230, opened 2025-01-17
feat: introduce HistoryLibraryBridge

### LINKED ISSUE #228: Introduce JNI bridge classes for HistoryLibrary (replacing `Bytes` with `byte[]`)
**Background**
Both the interfaces defined in the `hedera-services` need to have corresponding JNI bridge classes implemented, so that the interface implementations can call the bridges to call into the Rust code.

**Acceptance criteria**
Implement the JNI bridge classes (w/o implementing any native Rust methods just yet.)

anthony-swirldslabs:
**Description**:
Introducing the `HistoryLibraryBridge` class which is a JNI bridge to support an implementation of the `HistoryLibrary` defined at https://github.com/hashgraph/hedera-services/blob/main/hedera-node/hedera-app/src/main/java/com/hedera/node/app/history/HistoryLibrary.java

Similar to a previous PR at https://github.com/hashgraph/hedera-cryptography/pull/227 , this new bridge matches the interface in the services repository up to the `Bytes`/`byte[]` level. It also replaces the original `Map<Long, Bytes> sourceSignatures` with two arrays for performance reasons. Other than that, it's the same interface that's able to support the future implementation of the original services interface in full.

In the `cryptography/hedera-cryptography-hinTS/build.gradle.kts`, I'm updating the letter case of the native library name to match all the other usages as found in Cargo.toml and the HintsLibraryBridge class. This is to avoid any bizarre failures when attempting to load a native library with a wrong name.

**Related issue(s)**:

Fixes #228 

**Notes for reviewer**:
`gr build` succeeds

**Checklist**

### DIFF
--- a/cryptography/hedera-cryptography-hinTS/build.gradle.kts
+++ b/cryptography/hedera-cryptography-hinTS/build.gradle.kts
@@ -6,4 +6,4 @@ plugins {
     id("org.hiero.gradle.feature.test-multios")
 }
 
-cargo { libname = "hinTS" }
+cargo { libname = "hints" }
--- a/cryptography/hedera-cryptography-rpm/src/main/java/com/hedera/cryptography/rpm/Dummy.java
+++ b/cryptography/hedera-cryptography-rpm/src/main/java/com/hedera/cryptography/rpm/Dummy.java
@@ -1,4 +1,0 @@
-// SPDX-License-Identifier: Apache-2.0
-package com.hedera.cryptography.rpm;
-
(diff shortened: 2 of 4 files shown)


tinker-michaelj (approved):
LGTM, tyvm @anthony-swirldslabs !

andrewb1269 (approved):
Approve file `cryptography/hedera-cryptography-hinTS/build.gradle.kts`

mxtartaglia-sl (approved):

Issue #240, opened 2025-01-22
Handle deserialization failures


anthony-swirldslabs:
**Background**
If we pass garbage bytes from Java to JNI/hints level in Rust, trying to deserialize them into an object  is likely to produce:
```
thread '<unnamed>' panicked at src/main/rust/hints.rs:819:38:
called `Result::unwrap()` on an `Err` value: InvalidData
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
thread '<unnamed>' panicked at library/core/src/panicking.rs:221:5:
panic in a function that cannot unwind
```

This happens because the `deserialize` method just calls `unwrap()` unconditionally, trying to assert that it "knows what it's doing" and "guarantees the input bytes are good", which they may not be because they come from a Java app: https://github.com/hashgraph/hedera-cryptography/blob/a17801e277851417e7e35098def7064605c69f4a/cryptography/hedera-cryptography-hinTS/src/main/rust/hints.rs#L819-L819

While we trust the Java app, it would still be best to handle the error gracefully. With the current `panic` mode, we're unable to follow our contract and return `null` or `false`. Also, we're unable to create meaningful negative tests because the JVM crashes instead of returning from the JNI method with an error return code.

**Acceptance criteria**
* Remove the `unwrap()` call from the `deserialize()` implementation and make it return a `Result`
* Make callers responsible for processing the `Result` return value using `match ... {Ok/Err}` or similar.
* Return an error value from JNI method when a garbage provided as input.
* Create negative unit tests to verify the behavior of all the methods when garbage is given as input.

rsinha:
Agreed. IMO, a rule of thumb is to never panic inside the library, and return `Result`. In this case, we can define an Error type to contain deserialization error.

anthony-swirldslabs:
Apparently, the specific issue from the description of this ticket has been addressed a long time ago in a previous fix. However, there's still quite a number of unconditional `unwrap()` calls that could potentially cause a panic. So I'm hardening many of them as a part of https://github.com/hashgraph/hedera-cryptography/issues/277 in https://github.com/hashgraph/hedera-cryptography/pull/278 .

anthony-swirldslabs:
Subtasks of this ticket, as well as https://github.com/hashgraph/hedera-cryptography/pull/291 , finally resolved the issue in full.

PR #241, opened 2025-01-23
feat: implement client entropy in hinTS

rsinha:
**Description**: This PR allows the client of the hinTS library to supply all of the entropy needs, thus enabling the library to be pure functions. The expectation is on the platform to supply 256-bit random seeds where entropy is used.


rsinha:
@anthony-swirldslabs looks like this broke the build because `jni_crs` is using the older interface.

anthony-swirldslabs:
> @anthony-swirldslabs looks like this broke the build because `jni_crs` is using the older interface.

Right, the error looks like:
```
error[E0308]: mismatched types
  --> src/main/rust/jni_crs.rs:79:80
   |
79 |     let (crs, contribution_proof) = PowersOfTauProtocol::contribute(&prev_crs, random);
   |                                     -------------------------------            ^^^^^^ expected `[u8; 32]`, found `Fp<MontBackend<FrConfig, 4>, 4>`
   |                                     |
   |                                     arguments to this function are incorrect
   |
   = note: expected array `[u8; 32]`
             found struct `Fp<MontBackend<FrConfig, 4>, 4>`
note: associated function defined here
  --> src/main/rust/setup.rs:54:12
   |
54 |     pub fn contribute(crs: &CRS, seed: [u8; 32]) -> (CRS, ContributionProof) {
   |            ^^^^^^^^^^            --------------
```

I have a few comments about this:
1. It would be nice to use a `Vec<u8>` instead of an array as vectors seem to be easier to work with.
2. If that was a vector, then you can easily fix this failure by removing the `random` variable and passing the `random_vec` to the `contribute()` method. But if you want to continue to use an array, you'd have to find a way to convert the vector to the array that you want - I don't know an elegant way to do this. The approaches that I've seen don't look pretty, but maybe it's just because I don't know Rust.
3. I'm concerned about the fixed length of 32. Previously you suggested that the updateCRS/contribute method need 128 bits of randomness, which is 16 bytes. And that's what I'm using in my tests. While I'm not an expert in the Cryptography part of things, I think that different methods may want different amount of randomness. This is another reason why a vector would work better than a fixed size array. However, if the crypto algorithm does require 32 bytes of randomness every time, then you're going to have to modify the unit test for the updateCRS method to extend the `RANDOM` array to 32 values.

anthony-swirldslabs:
> 2. If that was a vector, then you can easily fix this failure by removing the `random` variable and passing the `random_vec` to the `contribute()` method. But if you want to continue to use an array, you'd have to find a way to convert the vector to the array that you want - I don't know an elegant way to do this. The approaches that I've seen don't look pretty, but maybe it's just because I don't know Rust.

Okay, https://stackoverflow.com/questions/29570607/is-there-a-good-way-to-convert-a-vect-to-an-array/29570662#29570662 seems to suggest that `vector.try_into().unwrap_or_else()` should do the trick, although it's a bit mouthful for my liking, but not as bad I guess. But again, as long as we use arrays, we have to fix the length, or deal with a variable length somehow (see my point 3 in the original reply above.)

So I suggest we figure out the length thing first, and then see how to go from vectors to arrays (and possibly error out if the length isn't right.)

rsinha:
I can change it to `impl AsRef<[u8]>` and that would accept all of the above. I will do a length check within the library.
As far as how much entropy, some places can live with 128 bits while others need 256 bits -- always passing 256 bits doesnt hurt, and is definitely safe.

rsinha:
BTW I noticed that the underly `rand_chacha` library expects [u8; 32]. So we would have to bite the bullet somewhere. What is your preference?

anthony-swirldslabs:
The length requirement of 32 sounds good to me.

As for where to convert, previously we agreed on a contract where the Java methods exported by the library return `null` or `false` in case of errors. So we need to be able to detect the error state in the JNI wrapper implementations (e.g. in that jni_crs.rs) to be able to return an error value per this contract. (On a side note, I had to file https://github.com/hashgraph/hedera-cryptography/issues/240 because currently it presents a problem for handling deserialization errors properly - please see the description for more details.)

So I suppose we'd have to do this in the JNI wrapper methods, something like:

```
let random_arr :[u8; 32] = random_vec.try_into()
        .unwrap_or_else(|v: Vec<T>| return std::ptr::null_mut());
```

(the above may be an invalid Rust code, but semantically it should do what's written there.)

And then we simply pass the random_arr to the hints/CRS library methods directly.

anthony-swirldslabs:
> let random_arr :[u8; 32] = random_vec.try_into()
>         .unwrap_or_else(|v: Vec<T>| return std::ptr::null_mut());

I suppose this should rather look something like this to be valid:
```
let random_arr :[u8; 32] = match random_vec.try_into() {
    Ok(val) => val,
    Err(_) => return std::ptr::null_mut() // or Err(())
}
```

cryptography/hedera-cryptography-hinTS/src/main/rust/hints.rs
```diff
@@ -187,9 +190,10 @@ pub struct VerificationKey {
 pub struct HinTS;
 
 impl HinTS {
-    /// generates a random secret key using the supplied random number generator
-    pub fn keygen<R: rand::Rng>(rng: &mut R) -> SecretKey {
-        F::rand(rng)
+    /// generates a random secret key using a PRNG seeded by the input entropy
+    pub fn keygen(seed: [u8; 32]) -> SecretKey {
```
anthony-swirldslabs, line 194: Minor: you may want to use the new `RANDOM_SIZE` constant everywhere where you hard-code `32` in this fix, so that it's defined in one place and can be updated at once if/when needed.

anthony-swirldslabs (approved):
LGTM

PR #243, opened 2025-01-27
chore: changes to HinTS api

rsinha:
**Description**: This PR makes modifications to the HinTS api to make it convenient for use. 
Specifically, we make the following changes:
- `partial_verify` works with extended public keys, thus avoiding extraction of public keys from extended public keys
- `verify_hint` accepts as input the signer's index and universe size


anthony-swirldslabs (approved):
LGTM. Thank you!

Issue #218, opened 2025-01-15
Write Rust methods to implement the `HintsLibraryBridge` and call the T2 Rust code (or use stubs until the T2 code is available)


anthony-swirldslabs:

anthony-swirldslabs:
Re-opening as the previous PR addressed the issue only partially (by design.)

PR #245, opened 2025-01-28
feat: introducing WRAPS

rsinha:
**Description**: This PR implements the (weighted) roster attestation proof system (a.k.a. WRAPS) for attesting that the latest roster is a descendant of the genesis roster. The proof system is built on the Succint zkVM toolchain.

Fixes # https://github.com/hashgraph/hedera-cryptography/issues/218


cryptography/hedera-cryptography-rpm/Cargo.toml
```diff
@@ -1,23 +1,37 @@
...
+ed25519-dalek = { version = "2.1.1", features = ["serde", "rand_core"] }
+hex = "0.4.3"
+rand = "0.8.5"
+serde = { version = "1.0", features = ["derive"] }
+serde-big-array = "0.5.1"
+serde-hex = "0.1.0"
+sha2 = "0.10.8"
+smallvec = { version = "1.13", features = ["const_generics", "serde"] }
+sp1-helper = "3.2"
+# sp1-verifier = "3.2"
```
anthony-swirldslabs: Minor: if you intend to use this, please add a comment just above this line in code and briefly explain why it's commented out.

Otherwise, it's best to delete this line and not to commit any commented-out code so as to avoid any confusion.

cryptography/hedera-cryptography-rpm/src/main/rust/common/src/address_book.rs
```diff
@@ -0,0 +1,110 @@
...
+
+    pub fn new(verifying_keys: Vec<ed25519::VerifyingKey>, weights: Vec<u64>) -> Self {
+        assert!(
+            verifying_keys.len() <= MAXIMUM_VALIDATORS,
+            "Too many verifying keys"
+        );
+        assert!(
+            verifying_keys.len() == weights.len(),
+            "Different number of verifying keys and weights"
+        );
```
anthony-swirldslabs, line 48: Minor: I presume this will panic, so we'd have to validate the inputs before calling this constructor, so that we can return an error to the Java layer. For this reason, it would be nice to have this validation implemented in a dedicated static method returning a boolean. This constructor would call the method and panic if it returns false. The JNI wrapper would call the method and simply return an error return value to Java immediately, thus preventing the panic.

cryptography/hedera-cryptography-rpm/src/main/rust/common/src/address_book.rs
```diff
@@ -0,0 +1,110 @@
...
+/// Calculates the total weight of the validators in the current AB
+pub fn calculate_total_weight(ab: &AddressBook) -> u64 {
+    ab.iter().map(|abe| abe.weight).sum()
+}
+
+/// Calculates the cumulative weight of the validators in the current AB
+/// that have signed the next AB's hash
+pub fn calculate_signers_weight(ab: &AddressBook, signatures: &Signatures, message: &[u8]) -> u64 {
+    // NOTE: assumes that signatures are presented in the same order
+    //       as the validators in the current AB
```
anthony-swirldslabs, line 74: Minor: so everyone must provide a signature? We don't support a mode where only some validators provided signatures? If we should support such a mode, then perhaps using maps instead of arrays might be a better choice.

cryptography/hedera-cryptography-rpm/src/main/rust/common/src/ed25519.rs
```diff
@@ -0,0 +1,64 @@
...
+
+#[repr(transparent)]
+#[derive(Debug, Deref, Clone)]
+pub struct SigningKey(pub ed25519_dalek::SigningKey);
+
+impl SigningKey {
+    pub fn generate() -> Self {
+        let mut csprng = rand::rngs::OsRng;
+        let signing_key: ed25519_dalek::SigningKey =
+            ed25519_dalek::SigningKey::generate(&mut csprng);
```
anthony-swirldslabs: Minor: could we pass the entropy seed from the caller, so as to be able to write deterministic tests?

cryptography/hedera-cryptography-rpm/src/main/rust/raps/src/byteraps.rs
```diff
@@ -0,0 +1,321 @@
...
+                ),
+            );
+
+            prev_proof = next_proof;
+            prev_roster = next_roster;
+        }
+    }
+}
+
+// A fixture that can be used to test the verification of SP1 zkVM proofs inside Solidity.
```
anthony-swirldslabs: Minor: similar to a comment above, please add a comment explaining why we keep the commented out code in the repository. Otherwise, it's best to remove it if we don't plan to use it right away.

cryptography/hedera-cryptography-rpm/src/main/rust/raps/src/raps.rs
```diff
@@ -0,0 +1,275 @@
...
+                *prev_proof
+                    .expect("expected previous proof")
+                    .proof
+                    .try_as_compressed()
+                    .unwrap(),
+                vk.vk.clone(),
+            );
+        }
+
+        println!(
```
anthony-swirldslabs: Minor: I'm unsure if we should `println` from production code.

cryptography/hedera-cryptography-rpm/src/main/rust/raps/src/raps.rs
```diff
@@ -0,0 +1,275 @@
...
+        signers: &[bool; N],
+        message: &[u8],
+    ) -> [Option<Signature>; N] {
+        core::array::from_fn(|i| signers[i].then_some(signing_keys[i].sign(message)))
+    }
+
+    fn debug(proof: &SP1ProofWithPublicValues) {
+        let parsed_proof =
+            PublicValuesStruct::abi_decode(&proof.public_values.to_vec(), true).unwrap();
+        println!("------------ BEGIN Roster Attestation Proof ------------");
```
anthony-swirldslabs, line 267: To add to an above comment: it's totally fine to `println` from a method called `debug`. So this looks fine here. But not up there.

cryptography/hedera-cryptography-rpm/src/main/rust/common/src/address_book.rs
rsinha: Do you mean that the rust function can assume valid inputs, where lengths match for example?

cryptography/hedera-cryptography-rpm/src/main/rust/common/src/address_book.rs
rsinha: This assumes only a subset of signers are present. `Signatures` is defined to be `Vec<Option<Signature>>`. I agree that maps are better, but I am not sure if they natively support serialization of maps to bytes -- vectors make the serialization simpler. We need serialization for communicating between the host rust library and the zkvm.

cryptography/hedera-cryptography-rpm/src/main/rust/common/src/address_book.rs
anthony-swirldslabs: I suppose it shouldn't just assume it and should panic still. I just suggest to extract the validation into a separate method that simply returns a boolean instead of causing a panic. This would allow the quoted call to call it and then panic explicitly, and it would also allow the JNI wrapper to also call the validation method and simply return an error code back to Java w/o a panic.

cryptography/hedera-cryptography-rpm/src/main/rust/common/src/ed25519.rs
rsinha: Yes, agreed.

cryptography/hedera-cryptography-rpm/src/main/rust/common/src/address_book.rs
rsinha: Makes sense. However, for this case, since the check is quite straightforward, is it better to simply have JNI bridge enforce it and return error to Java if the condition is not met? This may keep the rust-side interface small.

cryptography/hedera-cryptography-rpm/src/main/rust/common/src/address_book.rs
anthony-swirldslabs: Sounds good. What is confusing is that it says that the order of the signatures should match the order of validators in the current AB. Suppose I have AB with [v1, v2, v3], and suppose only the v1 and v3 provided the signature. The signature from v2 is missing. So with the remark in the code, it's unclear if the array should have only 2 elements, or all 3, and if it's the latter, then what should be at index 1 then?

If it's okay to skip the signature from v2 and have the array of size 2 here in this case, I'd suggest to mention that.

However, I believe that the logic of the code below actually assumes that the array should have an element for every validator.... So I'm unsure.

rsinha: I see. I should write comments to explain this then. This function expects a vector of size `n` for AB of size `n`. Present signatures have `Some(sig)` and absent signatures are `None`. So in your example, the vector would be `[Some(sig1), None, Some(sig3)]`.

rsinha: BTW, there is a test that illustrates a sample usage of this interface.

anthony-swirldslabs: Awesome, thanks for clarifying this! It would be great if the comment here in code mentioned this `None` thing briefly.

anthony-swirldslabs (dismissed):
Thanks for addressing comments. The code looks good to me now, so I'm approving the fix.

We only need to get it to build in its entirety now, and then the code is ready to go. Thanks!

anthony-swirldslabs (dismissed):
Still looks good to me.

jjohannes:
Status on Rust build:
- I would advice to merge `main` once #252 is in to speed things up
- I pushed an version update `sp1-sdk = "4.0.1"` - this solves the [cross-compilation issue cause by that dependency](https://github.com/succinctlabs/sp1/issues/263#issuecomment-1965187389). Please check if that update can be done. Currently, the Rust code using that library no longer compiles.
- Then, everything builds except for the windows cross-compilation: `:hedera-cryptography-rpm:cargoBuildX86Windows`. It fails with the following:

```
warning: ring@0.17.8: Compiler family detection failed due to error: ToolNotFound: Failed to find tool. Is `clang-cl` installed?

error: failed to run custom build command for `ring v0.17.8`

Caused by:

warning: build failed, waiting for other jobs to finish...

The following warnings were emitted during compilation:

warning: sp1-core-machine@4.0.1: Compiler family detection failed due to error: ToolNotFound: Failed to find tool. Is `clang-cl` installed?

error: failed to run custom build command for `sp1-core-machine v4.0.1`
```

I do not really understand what the problem is. It seems that `ring` library has specific demands as it does not only contain Rust code. Maybe you have an idea what we can do about it. The windows cross-compilation is [configured here](https://github.com/hiero-ledger/hiero-gradle-conventions/blob/main/src/main/kotlin/org/hiero/gradle/tasks/CargoBuildTask.kt#L94-L115). If we need to set some additional ENV variable, you can set them in you local environment to test. If it can be fixed by that, I can then integrate the fix into the Gradle setup afterwards.

rsinha:
@jjohannes I can fix the code as there were some breaking api changes in upgrading `sp1` to 4.0.01.

anthony-swirldslabs (approved):
LGTM, thank you!

mishomihov00 (approved):
Review applies to `.github/workflows/build.yml`

andrewb1269 (approved):

PR #262, opened 2025-02-11
test: add more tests for CRS methods

### LINKED ISSUE #261: Add more CRS tests

anthony-swirldslabs:
**Description**:
Adding more comprehensive tests for CRS method in the `HintsLibraryBridge`.
Also, partially addressing https://github.com/hashgraph/hedera-cryptography/issues/240 by improving the deserialization method in the `jni_util.rs`, and using the new capability in the tests.

**Related issue(s)**:

Fixes #261 

**Notes for reviewer**:
All tests should pass.

**Checklist**

### DIFF
--- a/cryptography/hedera-cryptography-hinTS/src/main/java/com/hedera/cryptography/hints/HintsLibraryBridge.java
+++ b/cryptography/hedera-cryptography-hinTS/src/main/java/com/hedera/cryptography/hints/HintsLibraryBridge.java
@@ -55,10 +55,10 @@ public class HintsLibraryBridge {
         if (signersNum < 1) {
             return null;
         }
-        return _initCRS(signersNum);
+        return initCRSImpl(signersNum);
     }
 
-    private native byte[] _initCRS(long signersNum);
+    private native byte[] initCRSImpl(long signersNum);
 
     /**
      * Update a previous CRS object with a new contribution.
--- a/cryptography/hedera-cryptography-hinTS/src/main/java/com/hedera/cryptography/hints/HintsLibraryBridge.java
+++ b/cryptography/hedera-cryptography-hinTS/src/main/java/com/hedera/cryptography/hints/HintsLibraryBridge.java
@@ -63,7 +63,7 @@ public class HintsLibraryBridge {
     /**
      * Update a previous CRS object with a new contribution.
      * @param prevCRS the previous CRS object
-     * @param random the random contribution of 128 bits (16 bytes)
+     * @param random the random contribution of 256 bits (32 bytes)
      * @return the updated CRS object and a concatenated contribution proof for the update
      */
     public native byte[] updateCRS(final byte[] prevCRS, final byte[] random);
--- a/cryptography/hedera-cryptography-hinTS/src/main/java/com/hedera/cryptography/hints/HintsLibraryBridge.java
+++ b/cryptography/hedera-cryptography-hinTS/src/main/java/com/hedera/cryptography/hints/HintsLibraryBridge.java
@@ -39,11 +39,26 @@ public class HintsLibraryBridge {
     ///////////////////////////////////////////////////////////////////////////
 
     /**
-     * Generates a new CRS object for the given number of signers.
-     * @param signersNum the number of signers
+     * Generates a new CRS object for the given number of signers plus 1.
+     * <p>
+     * The provided signersNum argument must be greater than the maximum number of nodes
+     * that the network supports by at least 1. For example, if the network supports up to 1023 nodes,
+     * then the signersNum argument must be equal to at least 1024. It's mathematically
+     * okay to have the singersNum much larger than the current number of nodes in the network,
+     * however, this will result in a larger space required to store the CRS.
+     *
+     * @param signersNum the number of signers plus 1
      * @return the CRS object
      */
-    public native byte[] initCRS(long signersNum);
+    public byte[] initCRS(long signersNum) {
+        // Support a degenerate case of 0 signers, or a normal case with more signers. Otherwise, error out.
+        if (signersNum < 1) {
+            return null;
+        }
+        return _initCRS(signersNum);
+    }
+
+    private native byte[] _initCRS(long signersNum);
 
     /**
      * Update a previous CRS object with a new contribution.
--- a/cryptography/hedera-cryptography-hinTS/src/main/rust/jni_crs.rs
+++ b/cryptography/hedera-cryptography-hinTS/src/main/rust/jni_crs.rs
@@ -29,7 +29,7 @@ use crate::jni_util;
 /// # Returns
 /// *   a byte array with a serialized CRS object, or null on error
 #[no_mangle]
-pub extern "system" fn Java_com_hedera_cryptography_hints_HintsLibraryBridge__1initCRS(
+pub extern "system" fn Java_com_hedera_cryptography_hints_HintsLibraryBridge_initCRSImpl(
     env: JNIEnv,
     _instance: JObject,
     signers_num: jlong,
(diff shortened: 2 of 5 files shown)


cryptography/hedera-cryptography-hinTS/src/test/java/com/hedera/cryptography/hints/HintsLibraryBridgeCRSTest.java
```diff
@@ -52,6 +69,24 @@ void testUpdateCRS() {
         updateAndVerifyCRS(prevCRS);
     }
 
+    @Test
+    void testUpdateCRSConstraints() {
+        assertNull(INSTANCE.updateCRS(null, CRSConstants.RANDOM));
+        assertNull(INSTANCE.updateCRS(initAndVerifyCRS(), null));
+
+        // Only byte[32] is valid for entropy input:
```
rsinha, line 77: I didn't follow why byte[32] specifically. More generally, what are the following assertions capturing?

cryptography/hedera-cryptography-hinTS/src/test/java/com/hedera/cryptography/hints/HintsLibraryBridgeCRSTest.java
```diff
@@ -62,5 +97,44 @@ void testVerifyCRS() {
                 Arrays.copyOfRange(nextCRSandContributionProof, CRS_4_LENGTH, nextCRSandContributionProof.length);
 
         assertTrue(INSTANCE.verifyCRS(prevCRS, nextCRS, contributionProof));
+
```
rsinha, line 100: You could also consider damaging by swapping some element in the vector with another -- that said, it might be difficult to do once you have serialized the CRS.

rsinha: By modifying some bytes in the CRS, we are most likely only triggering deserialization failures. If that is the intention of these tests, then it's perfectly fine.

cryptography/hedera-cryptography-hinTS/src/main/rust/jni_crs.rs
```diff
@@ -29,7 +29,7 @@ use crate::jni_util;
 /// # Returns
 /// *   a byte array with a serialized CRS object, or null on error
 #[no_mangle]
-pub extern "system" fn Java_com_hedera_cryptography_hints_HintsLibraryBridge_initCRS(
+pub extern "system" fn Java_com_hedera_cryptography_hints_HintsLibraryBridge__1initCRS(
```
mxtartaglia-sl: Is the name change intentional? Seems that there is an extra 1 in the name

mxtartaglia-sl: Never mind, I see its JNI convention for anything starting with _

cryptography/hedera-cryptography-hinTS/src/main/java/com/hedera/cryptography/hints/HintsLibraryBridge.java
```diff
@@ -39,11 +39,26 @@ public static HintsLibraryBridge getInstance() {
...
-    public native byte[] initCRS(long signersNum);
+    public byte[] initCRS(long signersNum) {
+        // Support a degenerate case of 0 signers, or a normal case with more signers. Otherwise, error out.
+        if (signersNum < 1) {
+            return null;
+        }
+        return _initCRS(signersNum);
+    }
+
+    private native byte[] _initCRS(long signersNum);
```
mxtartaglia-sl: I understand the reasoning for having the name starting with _, but it goes against standard Java's code style. I don't think we should blindly follow conventions, but I do think we have to be consistent in all codebase, and this would be one of the few locations we name private methods with _ (to the best of my knowledge).
My proposal: public InitiateCRS vs private initCRS

anthony-swirldslabs: Fair. I'm renaming `_initCRS` to `initCRSImpl`.

cryptography/hedera-cryptography-hinTS/src/test/java/com/hedera/cryptography/hints/HintsLibraryBridgeCRSTest.java
anthony-swirldslabs: I recall we discussed the random input/entropy in the past and agreed that we want to accept 256 bits of entropy for random input, which is equivalent to 32 bytes. This has been codified at: https://github.com/hashgraph/hedera-cryptography/blob/67fb03c6602864927f03efc8ca32d1172da89c4e/cryptography/hedera-cryptography-hinTS/src/main/rust/hints.rs#L41-L41

Therefore, the deserialization for the random input in native code uses a [u8; RANDOM_SIZE] array and requires the input size to match that.

Hence all these assertions.

cryptography/hedera-cryptography-hinTS/src/test/java/com/hedera/cryptography/hints/HintsLibraryBridgeCRSTest.java
anthony-swirldslabs: The CRS is defined at https://github.com/hashgraph/hedera-cryptography/blob/67fb03c6602864927f03efc8ca32d1172da89c4e/cryptography/hedera-cryptography-hinTS/src/main/rust/kzg.rs#L35-L41

as a struct containing two vectors. While I don't know exactly where these vectors are in the serialized form, there's a possibility to break the serialization per se (e.g. if the damaged bytes represent the length of the vector), but there's also a possibility to damage the data elements only, which must be (I presume) numbers. I'd love to be able to damage only the numbers, so that the deserialization succeeds and we check if the actual math in our implementation detects the damage or not. But w/o knowing the exact format of the serialized data, it's a bit difficult to do that deterministically. So my best bet was to modify some random elements.

If you have a better idea, or can point to specific indices to modify, I'd be more than happy to do that. Please let me know.

rsinha: On second thought, it is better to have those tests be done on rust side.

rsinha:
lgtm

cryptography/hedera-cryptography-hinTS/src/test/java/com/hedera/cryptography/hints/HintsLibraryBridgeCRSTest.java
anthony-swirldslabs: I disagree. There's multiple benefits of having them in Java. We're building an official Java interface for this library (in the form of the Bridge classes), and we need to ensure the interfaces are well tested, including any edge cases. By having the tests written in Java, we test the entire stack of the library - from the input interface, through the JNI bindings, down to the actual native code that implements the functionality. This way we don't need to test the same corner case multiple times at multiple levels - that wouldn't be extremely bad, of course, but would still be wasteful from LOC/effort perspectives.

Additionally, since 99% of the consensus node is implemented in Java, it's easier to implement these tests in Java as we have a lot more expertise in this language than any other. So it will be easier to maintain these tests in the future. At some point, we may even decide to port (parts of) the native code to Java, in which case we'd have all the tests already in place, and they could play a role of regression tests in this case to ensure the ported code is 100% compatible with the original native one.

mxtartaglia-sl (approved):

andrewb1269 (approved):

PR #267, opened 2025-02-12
feat: implement HistoryLibraryBridge

### LINKED ISSUE #231: Write Rust methods to implement the `HistoryLibraryBridge` and call the T2 Rust code (or use stubs until the T2 code is available)

anthony-swirldslabs:
**Description**:
Introducing an implementation for the `HistoryLibraryBridge`. I'm also enabling the previously introduced tests for the bridge and modifying the test constants to account for the latest changes in the RAPS library implementation in Rust.

Please note that I'm bypassing the `byteraps.rs` API and calling the `raps.rs` APIs directly from the bridge. This is done to avoid an extra layer in the calls to the actual Rust implementation. This approach is consistent with what we do in the hinTS library. Also, this simplifies errors handling as we need to write the `match/Err` only once - in the bridge methods themselves - instead of having to duplicate the exact same error handling in the `byteraps.rs` as well. For this reason, we may remove the `byteraps.rs` file altogether because it shouldn't be needed in this library. However, we can do that in a separate fix, perhaps when we remove the T1 implementation in the future.

**Related issue(s)**:

Fixes #231 

**Notes for reviewer**:
All tests should pass.

**Checklist**


rsinha (approved):

rbarker-dev (approved):
LGTM

PR #270, opened 2025-02-18
test: add more tests for HintsLibraryBridge

### LINKED ISSUE #264: Add more hinTS tests

anthony-swirldslabs:
**Description**:
Implementing constraints checks for `HintsLibraryBridge` methods and add extra tests to verify them, as well as verify the Rust code behavior when input data structures are corrupt.

**Related issue(s)**:

Fixes #264 

**Notes for reviewer**:
All tests should pass.

**Checklist**


tinker-michaelj (approved):
LGTM, tyvm @anthony-swirldslabs !

cryptography/hedera-cryptography-hinTS/src/main/java/com/hedera/cryptography/hints/HintsLibraryBridge.java
```diff
@@ -97,7 +97,14 @@ public byte[] initCRS(long signersNum) {
      * @param n the number of parties
      * @return the hints
      */
-    public native byte[] computeHints(final byte[] crs, final byte[] secretKey, int partyId, int n);
+    public byte[] computeHints(final byte[] crs, final byte[] secretKey, int partyId, int n) {
+        if (!validateCRS(crs, n) || !validatePartyId(partyId, n) || secretKey == null) {
```
rsinha, line 101: Is the CRS validation necessary? The (trusted) caller will be providing the CRS from state.

cryptography/hedera-cryptography-hinTS/src/main/java/com/hedera/cryptography/hints/HintsLibraryBridge.java
```diff
@@ -145,7 +181,14 @@ public native AggregationAndVerificationKeys preprocess(
      * @param privateKey the private key
      * @return the signature
      */
-    public native byte[] signBls(final byte[] message, final byte[] privateKey);
+    public byte[] signBls(final byte[] message, final byte[] privateKey) {
+        if (message == null || message.length == 0 || privateKey == null || privateKey.length == 0) {
```
rsinha, line 185: `privateKey` should have a fixed length, so we can assert something stronger.

cryptography/hedera-cryptography-hinTS/src/main/java/com/hedera/cryptography/hints/HintsLibraryBridge.java
rsinha: From below, I see that `validateCRS` is checking a property on the length of the string, rather than cryptographic verification. So perhaps this check is consistent with the other checks across other functions in this bridge.

anthony-swirldslabs:
@rsinha :
> Is the CRS validation necessary? The (trusted) caller will be providing the CRS from state.

I believe it is necessary for a couple of reasons:
1. It doesn't hurt performance-wise because the verification checks are very simple and should run really fast. At the same time, the checks may help prevent simple bugs now or in the future when we're integrating the library into the consensus nodes.
2. If the input arguments, including the CRS, are invalid, the native Rust code could panic in certain cases. It's best to catch such scenarios at the bridge level and return an error code gracefully rather than crash the JVM and stop the world.

What would be good reasons to remove the checks?

rsinha (approved):

cryptography/hedera-cryptography-hinTS/src/main/java/com/hedera/cryptography/hints/HintsLibraryBridge.java
anthony-swirldslabs: Yep. I further described this in a comment above as well.

cryptography/hedera-cryptography-hinTS/src/main/java/com/hedera/cryptography/hints/HintsLibraryBridge.java
anthony-swirldslabs: The length requirement is being enforced at the native Rust method already when we're trying to build an array of a fixed length (this happens indirectly via the error handling at `jni_utils::deserialize_jbyte_array`.) If that fails, the Rust method returns gracefully with `null`.

PR #278, opened 2025-02-26
chore: remove unconditional unwrap()

### LINKED ISSUE #277: Handle most failures

anthony-swirldslabs:
**Description**:
Removing unconditional calls to `unwrap()` to prevent potential panics in Rust and JVM crashes as a result. This is a sub-task of https://github.com/hashgraph/hedera-cryptography/issues/240 . Also, while at it, I went ahead and removed the `byteraps.rs` because it's unused.

Please note that there's dozens of `unwrap()` calls left in the `hints.rs` and surrounding files. I'm unsure how safe they are, but removing them and changing all the return value types to `Result` (or `Option`) in there would be a painfully large fix.

@rsinha what do you think? Is it worth doing in `hints.rs` and friends?

**Related issue(s)**:

Fixes #277 

**Notes for reviewer**:
All tests should still pass.

**Checklist**


cryptography/hedera-cryptography-rpm/src/main/rust/raps/src/raps.rs
```diff
@@ -124,28 +136,38 @@ impl RAPS {
...
-    });
+    let ab_prev_hash = match prev_proof {
+        Some(prev_proof) => {
+            let parsed_prev_proof = match PublicValuesStruct::abi_decode(&prev_proof.public_values.to_vec(), true) {
+                Ok(val) => val,
+                Err(_) => return Err(())
+            };
+            Some(parsed_prev_proof.ab_curr_hash.0)
+        },
+        None => None
```
rsinha, line 158: We can use `map_err` to simplify this pattern.

cryptography/hedera-cryptography-rpm/src/main/rust/raps/src/raps.rs
```diff
@@ -124,28 +136,38 @@ impl RAPS {
...
+    let tss_vk_prev_hash = match prev_proof {
+        Some(prev_proof) => {
+            let parsed_prev_proof = match PublicValuesStruct::abi_decode(&prev_proof.public_values.to_vec(), true) {
+                Ok(val) => val,
+                Err(_) => return Err(())
+            };
+            Some(parsed_prev_proof.tss_vk_hash.0)
+        },
+        None => None
+    };
```
rsinha, line 170: You can consider something as follows:
```
let tss_vk_prev_hash = prev_proof.map(|proof| {
    PublicValuesStruct::abi_decode(&proof.public_values.to_vec(), true)
        .map(|parsed| parsed.tss_vk_hash.0)
        .map_err(|_| ())? 
});
```

cryptography/hedera-cryptography-rpm/src/main/rust/raps/src/raps.rs
```diff
@@ -124,28 +136,38 @@ impl RAPS {
...
+    ab_genesis_hash: [u8; HASH_LENGTH],
     prev_proof: Option<&SP1ProofWithPublicValues>,
     vk_digest: [u32; 8],
     ab_curr: &AddressBook,
     ab_next: &AddressBook,
     signatures: &Signatures,
-    tss_vk_next_hash: [u8; 32],
-) -> ([u8; 32], [u8; 32], Statement) {
+    tss_vk_next_hash: [u8; HASH_LENGTH],
+) -> Result<([u8; 32], [u8; 32], Statement), ()> {
```
rsinha, line 146: If we are using the `Result` type, it might be worth defining an error enum with different possible error types across all the methods. We can return that instead of `()`.

cryptography/hedera-cryptography-rpm/src/main/rust/raps/src/raps.rs
anthony-swirldslabs: I'd love to use this, but it doesn't work because the `map_err` call happens inside an outer `map` closure. From the inside, whatever I `return`, becomes an output of the `map` closure, rather than a return value of the outer method itself. Here's an error message that I get:

```
error[E0277]: the `?` operator can only be used in a closure that returns `Result` or `Option` (or another type that implements `FromResidual`)
   --> src/main/rust/raps/src/raps.rs:153:29
    |
150 |     let ab_prev_hash = prev_proof.map(|proof| {
    |                                       ------- this function should return `Result` or `Option` to accept `?`
...
153 |             .map_err(|_| ())?
    |                             ^ cannot use the `?` operator in a closure that returns `[u8; 32]`
    |
    = help: the trait `FromResidual<Result<Infallible, ()>>` is not implemented for `[u8; 32]`

error[E0277]: the `?` operator can only be used in a closure that returns `Result` or `Option` (or another type that implements `FromResidual`)
   --> src/main/rust/raps/src/raps.rs:169:29
    |
166 |     let tss_vk_prev_hash = prev_proof.map(|proof| {
    |                                           ------- this function should return `Result` or `Option` to accept `?`
...
169 |             .map_err(|_| ())?
    |                             ^ cannot use the `?` operator in a closure that returns `[u8; 32]`
    |
    = help: the trait `FromResidual<Result<Infallible, ()>>` is not implemented for `[u8; 32]`
```

I don't really see an easy way to resolve this. One would guess that chaining `map` calls would solve this, but it won't because the error types of the results differ between the outer method and the inner calls.

So unfortunately, w/o any extravagant macros that may be difficult to maintain, I don't really see a way to simplify it.

cryptography/hedera-cryptography-rpm/src/main/rust/raps/src/raps.rs
anthony-swirldslabs: Please see below.

cryptography/hedera-cryptography-rpm/src/main/rust/raps/src/raps.rs
anthony-swirldslabs: Previously we agreed to return `null`/`false` on errors to Java. So there's not an immediate benefit from having descriptive errors because we won't be able to communicate them in a reasonable way to the Java API (w/o changing the API, that is.)

Further, this would complicate both the JNI bridge implementation, as well as the code in services that uses the library.

You can discuss this API change with services, and if they agree, we could implement this change. But for now, at least for this particular PR, I'd suggest to keep the thing as it is.

cryptography/hedera-cryptography-rpm/src/main/rust/raps/src/raps.rs
anthony-swirldslabs: I played with this more, and I still cannot find a solution that wouldn't look ugly or at least be shorter than what I currently have in this PR. It seems that there's not an elegant way to either transform an `Option<Result<T, E>>` into an `Option<T>`, or return a `Err` from the function if that inner Result is an error. Any approach I tried requires a huge if/else statement and doesn't really simplify the whole thing a lot.

rsinha: I was able to get the following to work:

```
let tss_vk_prev_hash = prev_proof
        .and_then(|proof| PublicValuesStruct::abi_decode(&proof.public_values.to_vec(), true).ok())
        .map(|parsed| parsed.tss_vk_hash.0);
```

I believe this is closer to idiomatic Rust.

cryptography/hedera-cryptography-rpm/src/main/rust/raps/src/raps.rs
rsinha: This should work:
```
let ab_prev_hash = prev_proof
        .and_then(|proof| PublicValuesStruct::abi_decode(&proof.public_values.to_vec(), true).ok())
        .map(|parsed| parsed.ab_curr_hash.0);
```

cryptography/hedera-cryptography-rpm/src/main/rust/raps/src/raps.rs
anthony-swirldslabs: It depends on the definition of "work". This does look shorter than what I have in my PR right now. And the code will indeed run fine provided all the inputs are correct. Then again, the code looks almost as short as the code that I'm changing here. And the reason for this is that your proposed solution doesn't handle the failure. Specifically, the `ok()` call will panic if the `Result` returned by `abi_decode` isn't actually okay. What I'm doing in this PR is I'm trying to eliminate the possibility of panics, including when processing the result of the inner `abi_decode` call.

So from the problem statement of this particular PR, your proposed code doesn't really work, unfortunately.

rsinha: I am not sure if `ok()` panics -- see https://github.com/rust-lang/rust/blob/00f245915b0c7839d42c26f9628220c4f1b93bf6/library/core/src/result.rs#L658. That said, I am fine with accepting this as well.

anthony-swirldslabs: Oh, my bad, I should've checked the spec for the `Result.ok()`. Yes, I agree it won't panic. However, it will silently swallow the failure and will result in an empty option for the `tss_vk_prev_hash`. This means that the code that uses the `Statement` produced by this function will now have to deal with the missing hash. If all stars align, that code could fail gracefully and return an error code. If they don't, that code could panic, w/o having any idea why this hash is missing because all the other logic in that code would suggest that the hash should be present, so that code has a total right to call `unwrap()` unconditionally on this hash, yet this will produce a panic...

The current version of this PR tries to detect the failure and fail fast by not returning a Statement which is by definition would be incorrect given the failure, allowing the caller to process this situation correctly and error out gracefully with certainty.

This seems to be the right thing to do given the spirit of this PR, so I'm a little reluctant to swallow this error condition as you suggest and let the caller deal with it however it can (or cannot.) But I can accept this if you insist.

Please let me know.

rsinha: I see. In that case, I agree with your change.

rsinha (approved):

PR #290, opened 2025-03-05
feat: reduce the native library size

### LINKED ISSUE #289: Reduce native lib size

anthony-swirldslabs:
**Description**:
Current builds produce `libraps.dylib` on Mac that is some 68MB in size, which is too big. The library is at least twice as small on other platform. This fix aims to reduce the size of the native library. With this PR, a Gradle build will result in a 18MB library instead.

This fix is inspired by https://users.rust-lang.org/t/rust-creates-huge-files/81884/6

**Related issue(s)**:

Fixes #289 

**Notes for reviewer**:
All tests should still pass.

**Checklist**


rsinha:
@anthony-swirldslabs Can we please see if `opt-level = z` makes things slower than 3? Given that the prover is slow, it might be worth having slightly larger binaries.

We may be getting most of the wins from stripping debug symbols, so perhaps opt-level 3 suffices?

anthony-swirldslabs:
> @anthony-swirldslabs Can we please see if `opt-level = z` makes things slower than 3? Given that the prover is slow, it might be worth having slightly larger binaries.
> 
> We may be getting most of the wins from stripping debug symbols, so perhaps opt-level 3 suffices?

Agree. opt-level 3 keeps the performance on par with before. With "z", the test was too very slow. So I'm updating the fix.

Please note that the library now produces a 22MB binary, which is still a huge improvement compared to the old 68MB.

rsinha (dismissed):

andrewb1269 (approved):
Approve version file.

rsinha (approved):

Issue #282, opened 2025-02-28
Remove panics from the hinTS library


rsinha:
We would like to avoid using asserts or unwraps, which can panic and shut down the process. Instead, we will return a `Result` type, and let the caller (e.g. the JNI bridge) decide how to handle it (e.g. forward error codes or simply return the appropriate result such as null).

rsinha:
Closed by #291

PR #291, opened 2025-03-06
feat: removing panics in hinTS

rsinha:
**Description**:
All hinTS methods have been modified to return a `Result` type rather than panicing, which aborts the process.

**Related issue(s)**: #282


anthony-swirldslabs (approved):
Changes in original commits from @rsinha look good to me. Approved.

rsinha:
Changes from @anthony-swirldslabs look good as well.

PR #296, opened 2025-03-07
fix: update CRS and n validation

### LINKED ISSUE #295: Update CRS and n validation constraint

anthony-swirldslabs:
**Description**:
Per a discussion with @Neeharika-Sompalli and @rsinha , we determined that the existing validation for CRS length and `n` is incorrect. This fix updates it making it possible to pass values of `n` that are still positive powers-of-two but may be smaller than the `signersNum` with which the CRS was initially generated. This is to allow for optimized computations in smaller networks which can still re-use a CRS generated for a potentially very large network size.

**Related issue(s)**:

Fixes #295

**Notes for reviewer**:
Also updating a test to verify the new constraint.

**Checklist**


Neeharika-Sompalli (approved):
LGTM! Thanks @anthony-swirldslabs

rsinha (approved):

PR #315, opened 2025-04-15
feat: put native library files into distinct Java packages

jjohannes:
**Description**:

This is required to load Java Modules that contain native parts on the Module Path without running into a split-package error like:

```
Error occurred during initialization of boot layer
java.lang.LayerInstantiationException: Package software.darwin.amd64 in both module com.hedera.cryptography.rpm and module com.hedera.cryptography.hints
```

Currently, `hedera-common-nativesupport`  expects all native libraries to be in the package `software.<os>.<arch>` . This has two issue:

1. It does not work to have native code in multiple modules, but in the same package (that's what I stumbled over)
2. We should not use the very generic package `software`. Instead these files should be in `com.hedera.cryptography` as the Java classes.

This PR to addresses these issues by now using the package `com.hedera.nativelib.<libname>.<os>.<arch>`.


.gitignore
```diff
@@ -69,23 +69,6 @@ gradle-app.setting
...
-
-# Test binary, built with `go test -c`
-*.test
-
-# Output of the go coverage tool, specifically when used with LiteIDE
-*.out
-
-# Dependency directories (remove the comment below to include it)
-# vendor/
-
```
jjohannes, line 88: Removed: no Go code in this repo and we put some binaries into this repo as test data - `libgreeter.*`.

anthony-swirldslabs (dismissed):

PavelSBorisov (approved):

anthony-swirldslabs (approved):
LGTM. Thank you!

andrewb1269 (approved):
Approve the `gitignore` and `settings.gradle.kts`

PR #331, opened 2025-04-29
feat: compression for WRAPS proofs

rsinha:
**Description**:
We implement a compression mechanism to compress 1.5 MB WRAPS proofs down to 1 KB.


cryptography/hedera-cryptography-rpm/src/main/rust/raps_compression_zkvm_program/src/main.rs
```diff
@@ -0,0 +1,55 @@
...
+    address_book::serialize_and_digest_sha256,
+    sha256::digest_sha256,
+    statement::CompressedStatement,
+    PublicValuesStruct, CompressedPublicValuesStruct,
+};
+use alloy_sol_types::SolValue;
+
+pub fn main() {
+    // Get the passed statement
+    println!("cycle-tracker-start: parsing statement_in");
```
anthony-swirldslabs, line 20: Minor: just for my understanding, as we agreed previously, we want our Go/Rust executable to have a `main()` and to communicate via stdin/stdout.

I also see this `main()` in this file here at line 18, and I see that it has debug `println!` statements. It doesn't seem to read any inputs from `stdin` though, nor does it print the output to the `stdout`.

Do I understand correctly that this is a **different** `main()` which seems to be "zkvm"-specific? And we'll have another, actual `main()` that's going to implement the stdin/stdout protocol for the Java code to interact with it?

rsinha: Ah, this is just the zkvm program that verifies the uncompressed proof. I am yet to commit the binary programs that we discussed.

cryptography/hedera-cryptography-rpm/src/main/rust/compressor/src/main.rs
```diff
@@ -0,0 +1,50 @@
...
+    ).expect("failed to compress proof");
+
+    // serialize the compressed proof to a byte array
+    let mut compressed_proof_buf: Vec<u8> = Vec::new();
+    bincode::serialize_into(&mut compressed_proof_buf, &compressed_proof).expect("failed to serialize proof");
+
+    let stdout = io::stdout();
+    let mut handle = stdout.lock();
+
+    handle.write_all(&compressed_proof_buf)?; // Writes the bytes
```
anthony-swirldslabs: Per the protocol, we need a big endian 4-bytes integer length prefix for the output as well. The input and output data format for arrays is completely symmetric.

anthony-swirldslabs (commented):
The code looks good to me, but there's one critical comment regarding the output byte array format on stdout. Please have a look.

cryptography/hedera-cryptography-rpm/src/main/rust/compressor/src/main.rs
rsinha: Ah, my mistake. Will fix.

anthony-swirldslabs (dismissed):

cryptography/hedera-cryptography-rpm/src/main/rust/raps/src/raps.rs
```diff
@@ -125,6 +130,56 @@ impl RAPS {
 
         verification.is_ok()
     }
+
+    pub fn compress_rotation_proof(
```
anthony-swirldslabs: Could you please clarify why we have a `compress_rotation_proof` here in this file, and also we have a function with the same name in the "native compressor" at `cryptography/hedera-cryptography-rpm/src/main/rust/compressor/src/main.rs` above?

At first glance, they look kinda similar (although I didn't do a line-by-line comparison). 

Would it be possible to have only a single copy of this method and call it from both the places?

anthony-swirldslabs (dismissed):
Looks good to me overall, but I posted one question.

cryptography/hedera-cryptography-rpm/src/main/rust/raps/src/raps.rs
rsinha: That's a good catch. I can remove it from the RAPS library. We just need the verifier in RAPS.

rbarker-dev (approved):

anthony-swirldslabs (approved):
LGTM. Thank you!

PR #332, opened 2025-04-29
fix: removing use of CRS in partial signature verification

rsinha:
**Description**:
This PR removes the use of CRS in partial signature verification. This is to avoid the penalty of deserializing the CRS on every call, which can happen several times for each threshold signing process.

### DIFF
--- a/cryptography/hedera-cryptography-hinTS/src/main/rust/hints.rs
+++ b/cryptography/hedera-cryptography-hinTS/src/main/rust/hints.rs
@@ -489,20 +489,18 @@ impl HinTS {
 
     /// verifies the partial signature under the signer's public key
     pub fn partial_verify(
-        crs: &CRS,
         msg: &[u8],
         ak: &AggregationKey,
         party_id: usize,
         sig: &PartialSignature
     ) -> Result<bool, HinTSError> {
         let lhs = <Curve as Pairing>::pairing(ak.pks[party_id], hash_to_g2(msg)?);
-        let rhs = <Curve as Pairing>::pairing(crs.powers_of_g[0], sig);
+        let rhs = <Curve as Pairing>::pairing(G1AffinePoint::generator(), sig);
         Ok(lhs == rhs)
     }
 
     /// verifies the list of partial signatures from a list of signers
     pub fn partial_verify_batch(
-        crs: &CRS,
         msg: &[u8],
         ak: &AggregationKey,
         signer_ids: impl AsRef<[usize]>,
@@ -514,7 +512,7 @@ impl HinTS {
         let agg_sig = add::<G2AffinePoint>(signatures.as_ref().iter().map(|sig| sig.clone()));
 
         let lhs = <Curve as Pairing>::pairing(apk, hash_to_g2(msg)?);
-        let rhs = <Curve as Pairing>::pairing(crs.powers_of_g[0], agg_sig);
+        let rhs = <Curve as Pairing>::pairing(G1AffinePoint::generator(), agg_sig);
         Ok(lhs == rhs)
     }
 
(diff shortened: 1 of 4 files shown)


cryptography/hedera-cryptography-hinTS/src/test/java/com/hedera/cryptography/hints/HintsLibraryBridgeTest.java
```diff
@@ -305,44 +305,32 @@ void testVerifyBlsConstraints() {
...
-        assertFalse(INSTANCE.verifyBls(crs, signature, HintsConstants.RANDOM_2, keys.aggregationKey(), -1));
-        assertFalse(INSTANCE.verifyBls(crs, signature, HintsConstants.RANDOM_2, keys.aggregationKey(), 666));
+        assertFalse(INSTANCE.verifyBls(null, HintsConstants.RANDOM_2, keys.aggregationKey(), partyId));
+        assertFalse(INSTANCE.verifyBls(EMPTY, HintsConstants.RANDOM_2, keys.aggregationKey(), partyId));
+        assertFalse(INSTANCE.verifyBls(signature, null, keys.aggregationKey(), partyId));
+        assertFalse(INSTANCE.verifyBls(signature, EMPTY, keys.aggregationKey(), partyId));
+        assertFalse(INSTANCE.verifyBls(signature, HintsConstants.RANDOM_2, null, partyId));
+        assertFalse(INSTANCE.verifyBls(signature, HintsConstants.RANDOM_2, EMPTY, partyId));
+        // assertFalse(INSTANCE.verifyBls(signature, HintsConstants.RANDOM_2, keys.aggregationKey(), -1));
+        // assertFalse(INSTANCE.verifyBls(signature, HintsConstants.RANDOM_2, keys.aggregationKey(), 666));
```
anthony-swirldslabs: These two tests verify the behavior when the partyId is invalid. Previously, the Java code validated them against the given CRS and errored out accordingly. However, with the CRS removed, this validation is no longer possible in Java code. When uncommented, these calls cause a crash in native code.

We need to find a way to validate the partyId and uncomment these two tests before releasing this fix.

anthony-swirldslabs: Seems like @rsinha 's commit fixed the tests, so they've been uncommented. Resolving the conversation.

cryptography/hedera-cryptography-hinTS/src/main/rust/hints.rs
```diff
@@ -489,32 +494,44 @@ impl HinTS {
...
+        // party_id can only be between 0 and n-2, inclusive
+        if party_id >= ak.n - 1 { // usize ensures non-negative
+            return Err(HinTSError::InvalidInput(
+                format!("signer_id {} out of range", party_id))
+            );
+        }
+
         let lhs = <Curve as Pairing>::pairing(ak.pks[party_id], hash_to_g2(msg)?);
-        let rhs = <Curve as Pairing>::pairing(crs.powers_of_g[0], sig);
+        let rhs = <Curve as Pairing>::pairing(G1AffinePoint::generator(), sig);
```
anthony-swirldslabs, line 510: Minor: this seems to imply that we now rely on `G1AffinePoint::generator()` to be deterministic and stable over time. In the event this function changes in future versions and starts returning a different value, our code here would break apart.

Are there any guarantees that the function will remain stable? Since this comes from an external dependency, could we add a comment to where the dependency is defined to indicate that the current version that we use returns the result that we expect here, while future versions would need to be tested before blindly upgrading them? To this end, it would be nice to create and point to a dedicated small test that would check specifically this thing, so as to simplify the upgrade process in the future. In fact, it would be best to have this test run during the build and fail the build if this function returns an unexpected result. Comments in the test would explain what's going on and how to resolve this.

anthony-swirldslabs (approved):
Rust code changes look good to me. But I left one comment.

cryptography/hedera-cryptography-hinTS/src/main/rust/hints.rs
rsinha: That's a great point! Thankfully, the generators come from an IETF standard, and will not change. The only time we expect it to change is when we retire the BLS12-381 curve and choose a different curve. But that requires the CRS to change as well, so the using `G1AffinePoint::generator()` is still fine here.

anthony-swirldslabs: This sounds like a good guarantee then. Thanks for clarifying this!

PR #335, opened 2025-04-30
feat: introduce ProofCompressor

### LINKED ISSUE #333: Introduce ProofCompressor
Add a class that uses the `NativeBinary` to extract the compressor from the JAR and run it as a process, and interact with it via stdin/stdout.

anthony-swirldslabs:
**Description**:
Introducing a `ProofCompressor` class that uses the `NativeBinary` to extract the compressor from the JAR and run it as a process, and interact with it via stdin/stdout.

Please see javadoc and comments in the class for more details on the format of the data transmitted via stdin/stdout.

**Revision 2:**
* Introducing a generic `ProcessAPI` that allows us to implement any API using an external process, and then using it in the `ProofCompressor` for this new concrete API implementation.
* Making the API accept 3 arguments - the prover key, verifying key, and the uncompressed proof.

**Related issue(s)**:

Fixes #333 

**Notes for reviewer**:
This class requires the binary to be present. So we'll create unit tests for this class under a separate issue https://github.com/hashgraph/hedera-cryptography/issues/334 once the binary is available at a later time.

**Checklist**

### DIFF
--- a/common/hedera-common-nativesupport/src/main/java/com/hedera/common/nativesupport/ProcessAPI.java
+++ b/common/hedera-common-nativesupport/src/main/java/com/hedera/common/nativesupport/ProcessAPI.java
@@ -1,0 +1,87 @@
+// SPDX-License-Identifier: Apache-2.0
+package com.hedera.common.nativesupport;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.OutputStream;
+import java.nio.file.Path;
+import java.util.Map;
+
+/**
+ * An AutoCloseable native process runner that allows one to interact with it via stdin/stdout by means
+ * of sending and receiving byte arrays using public methods in this class.
+ */
+public class ProcessAPI implements AutoCloseable {
+    private final Process process;
+
+    /**
+     * Create a new ProcessAPI instance and start the process.
+     * The process should normally exit once it'd done its job - e.g. completed an API call.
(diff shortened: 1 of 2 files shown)


rsinha:
The inputs to the compression system comprises three byte arrays: 1) the uncompressed proof; 2) the compression zkVM prover key; 3) the other is the WRAPS zkvm verification key. We may also decide to invoke the uncompressed prover using this mechanism as well, which has several inputs as well. 
How do you think we should handle this? We will have to serialize all these inputs as one byte array.

rsinha:
> The inputs to the compression system comprises three byte arrays: 1) the uncompressed proof; 2) the compression zkVM prover key; 3) the other is the WRAPS zkvm verification key. We may also decide to invoke the uncompressed prover using this mechanism as well, which has several inputs as well. How do you think we should handle this? We will have to serialize all these inputs as one byte array.

On second thought, the prover key and verification keys are fixed values. So, we can launch the compressor so that it knows how to grab these values (not sure how). Then, only the proof needs to be serialized to the process from the jvm. 

Edit: looks like we have to do this as the prover key is 150 MB, and it doesnt make sense to serialize that. So, sending over a single byte array containing the proof makes sense.

rsinha (dismissed):

rsinha (approved):

PR #344, opened 2025-05-07
fix: limit maxSigners

### LINKED ISSUE #343: Impose a limit on `n`/`maxSigners`
The memory usage for a given `n` is quadratic. So even if we use `short` type for the value, we may still run out of memory.

We want instead have a check that limits the value to something like 1024 (should be a configurable constant defined somewhere centrally, so that we can change the limit later if needed.)

anthony-swirldslabs:
**Description**:
Since CRS memory usage is quadratic relative to the `maxSigners` value, we want to limit the latter to a reasonable value, say `1024`. The limit is introduced as a constant so we can modify it if need be.

In this fix I check `signersNum` and `n` against this limit where the API receives the values as parameters. In a couple of places I'm checking the `partyId` against this limit where the API doesn't accept `n` explicitly.

**Related issue(s)**:

Fixes #343

**Notes for reviewer**:
A new unit test is added.

**Checklist**

### DIFF
--- a/cryptography/hedera-cryptography-hinTS/src/main/java/com/hedera/cryptography/hints/HintsLibraryBridge.java
+++ b/cryptography/hedera-cryptography-hinTS/src/main/java/com/hedera/cryptography/hints/HintsLibraryBridge.java
@@ -13,7 +13,7 @@ public class HintsLibraryBridge {
             new SingletonLoader<>("hints", new HintsLibraryBridge());
 
     /** The max number of signers that we can support w/o running into OutOfMemory as the memory usage is quadratic. */
-    private static final short MAX_SIGNERS_NUM = (short) 1024;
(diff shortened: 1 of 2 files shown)


rsinha:
I think it would be ideal to set it to 1023, not 1024. That would allow us to use degree 1024 polynomials. Otherwise we would have to use degree 2048. Wdyt?
That said, this number need not be a power of 2 (minus one), as we are just imposing a bound.

anthony-swirldslabs:
> I think it would be ideal to set it to 1023, not 1024. That would allow us to use degree 1024 polynomials. Otherwise we would have to use degree 2048. Wdyt? That said, this number need not be a power of 2 (minus one), as we are just imposing a bound.

Makes sense! Updated in the latest commit. Please have a look.

rsinha (approved):

PR #349, opened 2025-05-12
fix: additional error checking in hinTS

rsinha:
**Description**:
This PR implements additional safety checks in hinTS. While they are not strictly needed, as the existing checks will ensure the erroneous situations are not triggered, it is nevertheless useful to be more thorough.


anthony-swirldslabs (approved):
LGTM. Thank you!

PR #351, opened 2025-05-13
fix: limit max weights sum

### LINKED ISSUE #350: Limit the weights sum
To prevent u64 overflow (and signed long overflow on Java side) we want to limit the sum of weights to 2^63.

anthony-swirldslabs:
**Description**:
Limit the maximum sum of weights to 2^63 which doesn't overflow Java signed `long`, and certainly doesn't overflow Rust `u64`.

**Related issue(s)**:

Fixes #350 

**Notes for reviewer**:
A new test is added.

**Checklist**

### DIFF
--- a/cryptography/hedera-cryptography-hinTS/src/main/java/com/hedera/cryptography/hints/HintsLibraryBridge.java
+++ b/cryptography/hedera-cryptography-hinTS/src/main/java/com/hedera/cryptography/hints/HintsLibraryBridge.java
@@ -15,7 +15,7 @@ public class HintsLibraryBridge {
     /** The max number of signers that we can support w/o running into OutOfMemory as the memory usage is quadratic. */
     private static final short MAX_SIGNERS_NUM = (short) 1023;
 
-    /** The max theoretical sum of weights all nodes together can have, which is 2^63 because we use signed long. */
+    /** The max theoretical sum of weights all nodes together can have, which is 2^63-1 because we use signed long. */
     private static final long MAX_SUM_OF_WEIGHTS = Long.MAX_VALUE;
 
     static {
@@ -364,6 +364,7 @@ public class HintsLibraryBridge {
         try {
             long sum = 0;
             for (int i = 0; i < weights.length; i++) {
+                // Math.addExact() throws ArithmeticException if the sum overflows
                 sum = Math.addExact(sum, weights[i]);
             }
             return sum >= 0L && sum <= MAX_SUM_OF_WEIGHTS;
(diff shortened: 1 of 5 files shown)


cryptography/hedera-cryptography-hinTS/src/main/java/com/hedera/cryptography/hints/HintsLibraryBridge.java
```diff
@@ -15,6 +15,9 @@ public class HintsLibraryBridge {
     /** The max number of signers that we can support w/o running into OutOfMemory as the memory usage is quadratic. */
     private static final short MAX_SIGNERS_NUM = (short) 1023;
 
+    /** The max theoretical sum of weights all nodes together can have, which is 2^63 because we use signed long. */
+    private static final long MAX_SUM_OF_WEIGHTS = Long.MAX_VALUE;
```
rsinha, line 19: I am not very familiar, but Long.MAX_VALUE equals 2^63 - 1. Is that correct? Whereas we allow the sum to be 2^63?

cryptography/hedera-cryptography-hinTS/src/main/java/com/hedera/cryptography/hints/HintsLibraryBridge.java
```diff
@@ -351,4 +358,17 @@ private static int inferNFromCRSLength(final byte[] crs) {
...
     }
+
+    /** Check if the sum of weights doesn't exceed MAX_SUM_OF_WEIGHTS. */
+    public static boolean validateWeightsSum(final long weights[]) {
+        try {
+            long sum = 0;
+            for (int i = 0; i < weights.length; i++) {
+                sum = Math.addExact(sum, weights[i]);
+            }
+            return sum >= 0L && sum <= MAX_SUM_OF_WEIGHTS;
```
rsinha, line 370: Do we want to check for overflow in each iteration of the loop?

cryptography/hedera-cryptography-hinTS/src/main/java/com/hedera/cryptography/hints/HintsLibraryBridge.java
anthony-swirldslabs: Correct, it's 2^63-1. This is the maximum positive long that the Consensus Node code could store in a `long`. I've updated the javadoc accordingly.

cryptography/hedera-cryptography-hinTS/src/main/java/com/hedera/cryptography/hints/HintsLibraryBridge.java
anthony-swirldslabs: We do. The addExact() would throw if the sum overflows. I've added a clarifying comment in code.

rsinha (approved):

andrewb1269 (approved):

PR #365, opened 2025-06-10
build: change publishing to use maven central publishing portal

jjohannes:
**Description**:

⚠️ Only merge after we made switched the publishing mechanism on the Maven Central side.

For more details see: https://github.com/hiero-ledger/hiero-consensus-node/pull/19567

**Related issue(s)**:

https://github.com/hiero-ledger/hiero-consensus-node/pull/19567


rbarker-dev (approved):
Nit: `# TODO remove -PpublishTestRelease=true for fully automated (non-snapshot) releases` Verify we are ready to remove this?

PR #381, opened 2025-07-17
build: compressor

### LINKED ISSUE #379: Build compressor
The compressor build is triggered from a custom Rust build script when Cargo executes.

Our current build system uses Gradle, that uses a special `cargo` task from Hiero build utils. The latter uses `cargo-zigbuild` and `zig` in order to cross-compile Rust code. However, when building the compressor with `zig`, and error occurs:

```
error: symbol _runtime.covctrs not attached to any (sub)section
              note: while parsing /var/folders/gt/n3fxywks2nnbjgmbhhvjh3v40000gp/T/rustcSH7o2S/libsp1_recursion_gnark_ffi-60105640fd0f2cf8.rlib(go.o)
```

This issue seems to be fixed in zig 0.14+ ( https://github.com/ziglang/zig/issues/20243 ), but the `cargo-zigbuild` doesn't fully support zig 0.14+ currently. The idea is to not use `zig` when building the compressor.

However, when native Rust `cargo` is run in the `compressor/` directory, then it errors out because it cannot find stdlib.h when compiling native bindings. This issue has been solved by specifying the include path as:

```
$ CPATH=`llvm-config --includedir`/c++/v1 cargo build --release
```

This command succeeds and produces the compressor binary. Note that specifying the `CPATH` for our current Gradle build doesn't fix the build, so for now we have to go with the native Cargo build, at least for the compressor.

Problems to be solved:
1. "Decouple" the compressor so that the Gradle build doesn't try to build it, yet it should be buildable using the native Cargo.
2. Check the output path. Our Gradle build writes artifacts into `target/<arch>/release/` directory. However, the Cargo build for the compressor puts the binary directly under `target/release/`. This may or may not be an issue for cross-compilation.
3. Add a Gradle task to actually call Cargo for the compressor and check if the full Gradle build produces all the binaries - e.g. the already existing libraps.dylib and the new `compressor` program.
4. Make Gradle copy the compressor into the JAR, same as it does for the `libraps.dylib`.
5. See how to apply the solution to other platforms - they may use something different from the `CPATH`, and also something different from the `llvm-config`.
6. Once every platform works on its own, figure out our cross-compilation build setup so that the compressor is built on all platforms and makes it into the JAR in proper locations.

anthony-swirldslabs:
**Description**:
**THIS PR IS OPEN FOR REVIEW NOW.**

The `compressor` is a program implemented in the `sp1` library. A part of this program is written as a Go library that the compressor build builds from within its custom Rust/Cargo build script ( https://github.com/succinctlabs/sp1/blob/429e95e00a51db1f3d7257e7db73c7fe0fd40801/crates/recursion/gnark-ffi/build.rs#L36-L50 )

`hedera-cryptography` uses cross-compilation in order to build native libraries for all the supported platforms - windows-x86_64, linux-x86_64, linux-aarch64, darwin-aarch64, and darwin-x86_64. Since the Go library is native, it also needs to use the cross-compilation, and so does the compressor build as a whole. However, just adding the `src/main/rust/compressor` as a member of the rpm workspace results in a build error because:
1. The compressor's outer Cargo build is unable to build all the Rust dependencies using Zig that we use today for cross-compilation.
2. The inner Go build lacks any configuration for the cross-compilation and either produces wrong artifacts or never succeeds in the first place.

As a part of resolving the build issue, we've introduced https://github.com/succinctlabs/sp1/pull/2418 in `sp1` itself in order to support configuring the inner Go build. However, it's not released in an official release yet. Further, the very latest versions of `sp1` (5.2.x) fail to build some of their newer dependencies with our setup in general even when the cross-compilation isn't involved. The latest version that works is sp1 5.1.0 which is already an upgrade compared to our current dependency on sp1 4.2.1. In order to support this, the patch from the PR above has been applied to a custom fork on top of the sp1 5.1.0 at https://github.com/anthony-swirldslabs/sp1/tree/go-build-config-5.1.0 . We'll be using this fork until we resolve build issues with newer versions of `sp1`, which is a separate, unrelated work. Also, we can move the forked branch to a different repository if necessary.

In this PR, I'm adding the `doFirst{}` and `doLast{}` actions to the existing `CargoBuildTask` in Gradle that builds the main `rpm` (aka `raps`) native library. The `doFirst` is only responsible for installing a proper Go SDK, while the `doLast` implements the actual build of the compressor and copies the produced binary into a directory from where the outer Gradle build picks it up and packages it into the JAR of the `rpm` library.

Please refer to extensive comments in the build scripts that provide more details on how exactly this is currently setup. In a nutshell:
1. We can build for all platforms in GitHub.
2. Local, developer builds are supported on macos (tested on aarch64) and linux-x86_64 (untested, other than indirectly in GitHub itself) hosts out of the box. Local builds on other host platforms likely won't work and would require an extra complexity to be added to the build script. Currently, there's not a business use-case to add this extra-complexity though, so we simply don't support them for now.
3. Java unit tests have been modified to support the newer version of sp1 because some constants have changed with the upgrade from 4.2.1 to 5.1.0.

Comments in code provide deeper details on how the setup works.

**Related issue(s)**:

Fixes #379

**Notes for reviewer**:
Build succeeds. Tests pass.

**Checklist**


cryptography/hedera-cryptography-rpm/Cargo.toml
```diff
@@ -19,9 +21,14 @@ serde-big-array = "0.5.1"
...
 smallvec = { version = "1.13", features = ["const_generics", "serde"] }
-sp1-sdk = { version = "4.1.0", default-features = false }
-sp1-zkvm = { version = "4.1.0" }
-sp1-verifier = { version = "4.1.0" }
+
+# sp1 5.2.1 fails to build some of its dependencies, so for now we stick to 5.1.0.
+# Also, we use our custom fork with https://github.com/succinctlabs/sp1/pull/2418 applied.
+# Once the PR is merged into sp1 and released officially, and the build issues of the sp1 dependencies are resolved,
+# then we can switch back to using official sp1 releases:
+sp1-sdk = { version = "5.1.0", git = "https://github.com/anthony-swirldslabs/sp1", branch = "go-build-config-5.1.0", default-features = false }
```
nathanklick: This external dependency hosted in a personal account should be moved to either the `hashgraph` or `swirldslabs` GH organizations as a fork. 

We should never depend on a repository in a personal account. @CMiville42 @dr20240304

nathanklick (changes requested):

cryptography/hedera-cryptography-rpm/Cargo.toml
anthony-swirldslabs: I totally agree, that's the reason why I mentioned in the description that we can move this branch. I'll work with you off-line to do that.

anthony-swirldslabs: @nathanklick : I've just updated links to the forked repo. Please have a look. Thanks!

nathanklick (approved):

rsinha (approved):

PR #396, opened 2025-08-17
feat: optimization in WRAPS to only process one third of signatures in zkVM

rsinha:
**Description**:
We don't need to process more than a third of signatures by weight. With this optimization, we should see roughly 3x improvement in uncompressed prover running time.

Note that there are possibly more optimized implementations for this change, but this logic is running outside of the prover, and is a minuscule fraction of the total running time -- so this PR is prioritizing simplicity.


cryptography/hedera-cryptography-rpm/src/main/rust/common/src/address_book.rs
```diff
@@ -95,6 +95,48 @@ pub fn calculate_signers_weight(ab: &AddressBook, signatures: &Signatures, messa
...
+    let total_weight: u64 = calculate_total_weight(ab);
+    // we want to be at least a third of that weight, rounded up to the next integer
+    let threshold_weight: u64 = total_weight / 3 + ((total_weight % 3) != 0) as u64;
+
+    // let us sort the address book weights, so we can trim them in order of increasing weight.
+    let mut weights: Vec<u64> = ab.iter().map(|abe| abe.weight).collect();
+    weights.sort_unstable();
+
+    let mut output_signatures = signatures.clone();
+    for &weight in weights.iter() {
```
anthony-swirldslabs, line 114: Minor: seems like we're pruning low-weight signatures. This algorithm may become inefficient if the size of the network grows a lot. On top of this, I see that below we clone() quite a bit as well as run inner loops, making the algorithm O(n^2), which further adds to the inefficiency in both memory and compute.

Perhaps a more efficient algorithm would perform a reverse sorting, and then build up a list of signatures starting from the heaviest weight until enough is accumulated.

anthony-swirldslabs (approved):
LGTM. Just a minor comment.

cryptography/hedera-cryptography-rpm/src/main/rust/common/src/address_book.rs
rsinha: That's a good point. I didn't think of that. Let me see if I can modify it.
But, I should mention that if n grows, we will face the bottleneck with prover time much earlier than this algorithm.

PR #431, opened 2025-10-09
feat: improved WRAPS construction

rsinha:
**Description**:
This PR introduces the new WRAPS construction, which uses folding-based SNARKs.

The advantages over the previous construction are:
- smaller running time: 2 minutes compared to about 90 minutes
- language dependencies: pure Rust as opposed Rust + Go
- software dependencies: 104 Rust crates, as opposed to approximately 900.
- memory consumption: roughly similar to before, but allows for further optimizations (in scope for future PRs).


cryptography/hedera-cryptography-WRAPS/src/main/java/module-info.java
```diff
@@ -0,0 +1,9 @@
+// SPDX-License-Identifier: Apache-2.0
+/**
+ */
+module com.hedera.cryptography.wraps {
+    exports com.hedera.cryptography.wraps;
+
+    requires com.hedera.common.nativesupport;
+// requires com.hedera.cryptography.hints;
```
edward-swirldslabs: Is this dependency no longer needed?  Do we need to add an explanation why the commented out dependency remains in the code?

edward-swirldslabs (dismissed):

cryptography/hedera-cryptography-WRAPS/src/main/java/module-info.java
anthony-swirldslabs: This is a WIP. The JNI bridge is being built as we speak. More commits are coming into this branch a bit later. 95% of changes in these future commits will be in the JNI bridge itself. However, a few changes will be present in the existing Rust native code (some extra utils and such.)

rbarker-dev (dismissed):
LGTM

Jeffrey-morgan34:
@rbarker-dev @edward-swirldslabs could we please get another review of this so we can try to get it into the 0.68 release?

anthony-swirldslabs (approved):
LGTM. Note that I don't possess enough expertise in cryptography and/or math to even barely understand what sort of computations are implemented in this code. Also, my expertise in Rust is rather limited. But I've read the code, and purely from the code perspective, I didn't find anything obviously wrong.

cryptography/hedera-cryptography-WRAPS/src/main/rust/wraps/src/lib.rs
```diff
@@ -0,0 +1,1219 @@
...
+        let padded_bitvector: [bool; MAX_AB_SIZE] = {
+            let mut vec = bitvector.as_ref().to_vec();
+            vec.resize(MAX_AB_SIZE, false);
+            vec.try_into().unwrap()
+        };
+
+        let is_genesis: bool = prev_proof.is_none();
+        if is_genesis {
+            // ensure genesis ab hash matches
+            assert_eq!(*ab_genesis_hash, hash_addressbook(&padded_prev_ab)?);
```
edward-swirldslabs, line 878: For every assert in this file, what happens if the assert is violated?  Do we have any tests that can cause these asserts to trigger?  Does it crash the JVM or do we handle it elegantly?

cryptography/hedera-cryptography-WRAPS/src/main/java/com/hedera/cryptography/wraps/WRAPSLibraryBridge.java
```diff
@@ -0,0 +1,392 @@
...
+        if (round2Messages == null) {
+            round2Messages = EMPTY_BYTE_ARRAY_2;
+        }
+        if (round3Messages == null) {
+            round3Messages = EMPTY_BYTE_ARRAY_2;
+        }
+
+        if (phase != SigningProtocolPhase.Aggregate) {
+            if (schnorrPrivateKey == null || instanceEntropy == null || instanceEntropy.length != ENTROPY_SIZE) {
+                return null;
```
edward-swirldslabs, line 132: Should these null returns be replaced with `IllegalArgumentException` explaining the nature of the error?  If failure is occurring, how do we differentiate the different ways of failing?  Or are we expecting to never fail?  Would it be helpful just in case we do fail to know the reason?

cryptography/hedera-cryptography-WRAPS/src/main/rust/wraps/src/lib.rs
```diff
@@ -0,0 +1,1219 @@
...
+        external_inputs: Self::ExternalInputsVar,
+    ) -> Result<Vec<FpVar<Fr>>, SynthesisError> {
+
+        let prev_pk_vars = (0..K)
+            .map(|i| JubJubVar::new_witness(cs.clone(), || Ok(
+                ark_ed_on_bn254::EdwardsAffine::new(
+                    external_inputs.0[3*i + 0].value()?,
+                    external_inputs.0[3*i + 1].value()?
+                )
+            )).unwrap())
```
edward-swirldslabs, line 252: There are `unwraps()` that don't have comments about their safety.  The AI agents I'm using for code review are indicating any unwrap that fails might trigger a panic and crash the JVM.   Is there any way we can do unit testing with inputs that try to hit this unwraps to prove they are safe?   Would it ever be possible that bad input can reach any of these unwraps?

cryptography/hedera-cryptography-WRAPS/src/main/rust/wraps/src/utils.rs
```diff
@@ -0,0 +1,70 @@
...
+
+/// Gets a ProvingKey.
+pub fn get_proving_key() -> Result<ProvingKey, WRAPSError> {
+    let artifacts_path = match env::var(ARTIFACTS_PATH_ENV_VAR) {
+        Ok(val) => val,
+        Err(_) => return Err(WRAPSError::BinaryArtifactMissing)
+    };
+
+    let nova_pp_path = artifacts_path.clone() + "/nova_pp.bin";
+    let nova_pp_bytes = match std::fs::read(nova_pp_path) {
```
edward-swirldslabs, line 36: nova read from disk.  Since there is no caching, and whenever we read it from disk there is a 2GB RAM spike, what's the frequency of our reading it from disk?  Since we have to eat the 2GB RAM at some point, if we hit a ceiling, wouldn't we hit it there?  If we read from disk frequently, can we cause thrashing?   Is the whole 2GB needed at once, or is it possible to stream it and use it chunks at a time?   Is there any change that two threads/processes invoke loading this file from disk in parallel at the same time?

cryptography/hedera-cryptography-WRAPS/src/main/rust/wraps/src/lib.rs
```diff
@@ -0,0 +1,1219 @@
...
+            instance
+        } else {
+            // Resume the incremental IVC from the previous proof.
+            let ivc_proof = NovaProof::deserialize_compressed(prev_proof.unwrap().as_slice()).unwrap();
+            N::from_ivc_proof(ivc_proof, (), (pk.nova_pp.clone(), vk.nova_vp.clone()))
+                .map_err(|_| WRAPSError::CryptographyError)?
+        };
+
+        // Fold in the current rotation step and immediately sanity-check the resulting IVC proof.
+        ivc_instance.prove_step(thread_rng(), VecF(external_inputs_at_step.clone()), None)
```
edward-swirldslabs, line 923: Are these `thread_rng()` calls truely random across different machines, or pseudo-random and every machine computes the same random number at the same point?  

What role is randomness playing here?

edward-swirldslabs (approved):

cryptography/hedera-cryptography-WRAPS/src/main/java/com/hedera/cryptography/wraps/WRAPSLibraryBridge.java
anthony-swirldslabs: No, at least not in this PR. Previously, when we were building JNI bridges for hinTS and RPM modules, we decided that the APIs are going to return `null` or `false` in case of any errors. IIRC, the performance and simplicity were the main arguments in favor of this approach as suggested by the Execution team. If this decision needs to be revisited, then it's outside of scope of this particular PR.

cryptography/hedera-cryptography-WRAPS/src/main/rust/wraps/src/utils.rs
anthony-swirldslabs: A comment above these methods provides some details on the memory usage and potential optional caching. As of today, it's unobvious if this method is going to be called more frequently than about once a month, or maybe once a day at most. Per the aforementioned comment, we may revisit the implementation when more data is available.

rsinha: Note that this is in a test module, which is why it has not been optimized any further. Release mode compilation will remove all of this code.

cryptography/hedera-cryptography-WRAPS/src/main/rust/wraps/src/lib.rs
rsinha: We are not relying on proofs being deterministic, so it is possible for different machines to compute different proofs. The randomness here is purely for adding zero-knowledge, which is only present here to maintain similarity with the Nova paper, and not required for our use case (as we don't need privacy).

cryptography/hedera-cryptography-WRAPS/src/main/rust/wraps/src/utils.rs
anthony-swirldslabs: Not really. These two methods are being invoked in the production code. E.g. see jni_wraps.rs at line 325.

rsinha: Ah I confused it with another thing. Thanks for the clarification.

cryptography/hedera-cryptography-WRAPS/src/main/rust/wraps/src/lib.rs
rsinha: I couldnt get rid of this unwrap due to struggles with the type system.  That said, this is executed during circuit setup (this method creates the circuit), and will not be executed during normal operation. 

Worth adding a comment at least, I agree.

cryptography/hedera-cryptography-WRAPS/src/main/rust/wraps/src/utils.rs
edward-swirldslabs: I'm fine with resolving this issue.  I can see that its loaded on the initialization of the prover.   Is that only in the genesis case? Or every proof step?

anthony-swirldslabs: It's for every proof. That's why I mentioned that it's going to be invoked either once a month or once a day - i.e. on every AddressBook rotation.

cryptography/hedera-cryptography-WRAPS/src/main/rust/wraps/src/lib.rs
rsinha: This is an example of a sanity check. If it fails, it means the caller (execution) is using the API incorrectly, indicating a software bug. To your question, it does crash the process, but not the JVM as the prover runs as a separate process.

anthony-swirldslabs: I'd like to clarify one point: there's a plan to make the prover run in a separate process eventually and provide this capability as a feature of this library. The current implementation does not provide such a capability. Unless the caller adds an extra complexity on their end to run the prover in a separate process, it will actually crash.

Personally, I'd prefer if the asserts were replaced with proper `return Err(...)` which the JNI bridge is able to handle gracefully w/o producing a crash. This way this code would be safe even if we don't have separate processes.

anthony-swirldslabs:
@nathanklick : Please note that I'm unable to use `feat!: ...` in the commit message for some reason. The `Title Check` errors out with `No release type found in pull request title` then:

https://github.com/hashgraph/hedera-cryptography/actions/runs/18916126835/job/53999884949?pr=431

However, we'd prefer to release this as `3.0.0`. Not sure how to do this. Please advise.

cryptography/hedera-cryptography-WRAPS/src/main/rust/wraps/src/lib.rs
edward-swirldslabs: I agree with Anthony.  I don't want to risk that we never get around to moving this code outside the JVM and we leave code in that has a chance of crashing the JVM.   This also applies to any unwraps that we can't guarantee will never fail as well.

rsinha: I agree as well. Let me make the change, here and other relevant locations.

rbarker-dev (approved):
LGTM

cryptography/hedera-cryptography-WRAPS/build.gradle.kts
nathanklick, line 1: The folder name for Gradle projects should be all lowercase. So `hedera-cryptography-WRAPS` needs to change to `hedera-cryptography-wraps` in order to be compliant.

README.md
```diff
@@ -11,6 +11,7 @@ This section is a work in progress.
 The repository includes the following projects:
 * **cryptography/hedera-cryptography-hinTS**: HinTS API that allows participants to calculate their hints, generate keys, and produce and verify aggregate signatures.
 * **cryptography/hedera-cryptography-rpm**: History API, aka RAPS or WRAPS, that allows participants to generate and verify recursive proofs for AddressBooks.
+* **cryptography/hedera-cryptography-WRAPS**: WRAPS 2.0 library that allows participants to generate and verify recursive proofs for AddressBooks.
```
nathanklick, line 14:
```suggestion
* **cryptography/hedera-cryptography-wraps**: WRAPS 2.0 library that allows participants to generate and verify recursive proofs for AddressBooks.
```

nathanklick (approved):

cryptography/hedera-cryptography-WRAPS/build.gradle.kts
anthony-swirldslabs: https://github.com/hashgraph/hedera-cryptography/issues/439

cryptography/hedera-cryptography-WRAPS/src/main/rust/wraps/src/lib.rs
anthony-swirldslabs: @rsinha : please file a ticket for this follow-up change.

README.md
anthony-swirldslabs: https://github.com/hashgraph/hedera-cryptography/issues/439

PR #436, opened 2025-10-24
build(deps): bump org.hiero.gradle.build to 0.6.1 / Gradle to 9.2.0

jjohannes:
**Description**:

- Update Gradle to 9.1
- [Update Gradle conventions to 0.6.0](https://github.com/hiero-ledger/hiero-gradle-conventions/releases/tag/v0.6.0)
  - remove `plugins` blocks from `gradle/aggregation` and `hiero-dependency-versions` as these are now configured by default through the conventions
  - Remove unused versions from `hiero-dependency-versions` as advised by new [checkVersionConsistency](https://github.com/hiero-ledger/hiero-gradle-conventions/issues/99) check 
  - Cleanup `gradle.properties`


gradle.properties
```diff
@@ -1,14 +1,6 @@
...
 
 # Enable Gradle Caching
 org.gradle.caching=true
 org.gradle.configuration-cache=true
-
-# Disable parallel workers
-# If/when we upgrade to cargo-zigbuild 0.20.0, it fails with parallel workers.
-# See https://github.com/rust-cross/cargo-zigbuild/issues/348
-# But even before the upgrade, builds seem to run faster this way.
-org.gradle.parallel=false
```
jjohannes, line 14: If `org.gradle.configuration-cache=true` (CC) then `org.gradle.parallel` has no effect, because CC allows for even [more parallelism in general](https://docs.gradle.org/current/userguide/configuration_cache.html#config_cache:intro).

Whatever was observed here must have been some (side) effect of something else.

If the cargo-zigbuild update is causing issues, we probably should change something in the `org.hiero.gradle.feature.rust` setup inside gradle-conventions. The linked issue from the comment seems to be fixed though.

andrewb1269 (dismissed):

andrewb1269 (approved):

PR #446, opened 2025-11-05
perf: adding a pruneCRS method to use a smaller CRS in all hinTS oper…

rsinha:
This PR introduces a method to prune the size of CRS. 
The method takes as input the existing CRS, and a new degree (typically a power of 2).


anthony-swirldslabs (approved):
LGTM. And I've added a JNI bridge method for the new API.

PR #404, opened 2025-08-19
test: download Groth16 artifact for ProofCompressorTest

anthony-swirldslabs:
**Description**:
In this PR, a new Gradle task is added to download the large (3GB) binary artifact required by the ProofCompressorTest to actually test the compressor. W/o the artifact the test was a no-op.

The artifact itself is available at:

https://builds.hedera.com/tss/sp1/groth16/v5.0/v5.0.0-groth16.tar.gz

per https://github.com/hiero-ledger/hiero-consensus-node/issues/20474#issuecomment-3192449577

**Related issue(s)**:

Fixes #391

**Notes for reviewer**:
All tests must pass.

**Checklist**

- [x] Documented (Code comments, README, etc.)
- [x] Tested (unit, integration, etc.)

PR #450, opened 2025-11-06
build: replace buildSrc with gradle/plugins

jjohannes:
**Description**:

This is an attempt to fix the currently broken _snapshot publish_ on `main`.

It looks like introducing `buildSrc` in #404 may be the cause. At least that's where it starts failing. Somehow between the `assemble` step and the `publishAggregationToCentralPortal` step the Gradle plugins classpath changes. Which leads to all custom tasks, like our rust compilation, being re-executed. This is very cryptically expressed by

```
The task was not up-to-date because of the following reasons:	
One or more additional actions for task ':hedera-cryptography-rpm:cargoBuildAarch64Darwin' have changed.
```

here:
https://scans.gradle.com/s/mjqnfhj5wf4nm/timeline?cacheability=cacheable&details=62vz3j4o3mc6y&hide-timeline&kind=task&outcome=success,failed&sort=longest
I assume that `buildSrc`, which creates an additional Jar for the plugins classpath, somehow causes this. Maybe the ordering changes between the two invocations or something.

`buildSrc` is a bit of a weird thing in current Gradle. It is supposed to behave as any "plugin build" - `pluginManagement { includeBuild("some-plugin-build") }` - but it isn't completely doing that as in early Gradle version the general concept of "plugin builds" did not exists. 

Moving to an individual plugin build may fix the issue.

This PR also introduces the `workflow_dispatch:` trigger for the snapshot publish job. If this PR should not fix the issue, we can run the workflow individually in follow up PRs for testing and debugging.

**Related issue(s)**:

#404


.github/workflows/publish_snapshot_release.yml
```diff
@@ -87,9 +88,8 @@ jobs:
         run: ./gradlew assemble
 
       # The steps below publish the artifacts to the Maven Central Nexus repository.
-      # TODO remove -PpublishTestRelease=true for fully automated (non-snapshot) releases
```
jjohannes, line 90: This is outdated. This workflow is never used for publishing a final release. That is done through `.releaserc`.
The `publishTestRelease` flag has no effect on snapshot publishing.

andrewb1269 (approved):

PR #458, opened 2025-11-11
feat: support optional memory-on-disk for WRAPS

### LINKED ISSUE #444: Custom GlobalAlloc backed by memory-mapped files
WRAPS 2.0 uses up to ~20GB of RAM when constructing an AddressBook rotation proof (measured by running an existing unit test that uses very small ABs. The memory usage may (or may not) be larger for larger ABs.)

To enable running this code on smaller systems with less RAM, the idea is to build a custom global memory allocator for the WRAPS 2.0 library that would use a memory-mapped file on disk. The operating system would handle allocation/deallocation and I/O as the process uses the memory.

This, of course, could and would result in the performance degradation of the proof construction method. Today, using real RAM, this operation takes about 200 seconds on a developer Mac laptop. However, this operation needs to run only when the AddressBook needs to rotate, which happens once a month in real networks today. This operation may happen more frequently in the future. However, currently we think that the proof construction may easily take an hour, and that would be fine. So we can try and trade the RAM for disk, as well as the performance for RAM.

anthony-swirldslabs:
**Description**:
WRAPS 2.0 may use up to 20 GB memory when constructing an AddressBook change proof. This is measured by the unit test ~which uses very small ABs. So I presume the code may require even more memory with real-sized ABs. Regardless, even~ the 20 GB may be too large for smaller environments such as Solo use-cases and similar. So we want to have an option to use a file on disk for memory used by the WRAPS 2.0 code.

Introducing a `MemmapAllocator` that implements a `GlobalAlloc` memory allocator in Rust for the WRAPS 2.0 native code. The allocator uses a memory-mapped file on disk to allocate memory for the library. See the head of the `alloc.rs` for constant size definitions. The allocator uses two utility objects:
* `FileMap` that holds a reference to the memory-mapped file itself
* `BitMap` that maintains a bit-map of (un)allocated blocks in the `FileMap`.

This feature is off by default and the library uses the system RAM by default. To enable the on-disk memory, one needs to specify the `TSS_LIB_WRAPS_SWAP_FILE` environment variable with a name of the file where that on-disk memory will be located. If any errors occur (or if we run out of the memory on-disk), then the allocator will fall back to using the system RAM again.

I've only been able to test this on my developer laptop which has a lot of free RAM. So even with this feature enabled and allocations actually happening in the memory-mapped file, the OS probably didn't touch the actual disk too much. Runtime-wise, running tests took pretty much the exact same time as without this custom allocator enabled. However, memory usage in the System Monitor clearly showed that the system RAM usage for the `java` process was almost always around 100MB and only briefly spiked to something like 6 GB, while the process kept using some 20 GB of the total memory - that was our on-disk file usage.

**Related issue(s)**:

Fixes #444

**Notes for reviewer**:
The feature is disabled by default, so there's no any breaking changes.

**Checklist**

### DIFF
--- a/cryptography/hedera-cryptography-wraps/build.gradle.kts
+++ b/cryptography/hedera-cryptography-wraps/build.gradle.kts
@@ -28,6 +28,10 @@ tasks.test {
                     .dir("v0.1.0")
                     .asFile
                     .absolutePath,
+
+            // Commented-out just to provide an example of how to enable swap for WRAPS 2.0.
+            // When not set, the proof construction may require up to ~20GB of RAM.
+            // "TSS_LIB_WRAPS_SWAP_FILE" to "/tmp/MemoryMapFile",
         )
     )
 }
--- a/cryptography/hedera-cryptography-wraps/src/main/java/com/hedera/cryptography/wraps/WRAPSLibraryBridge.java
+++ b/cryptography/hedera-cryptography-wraps/src/main/java/com/hedera/cryptography/wraps/WRAPSLibraryBridge.java
@@ -30,6 +30,13 @@ public class WRAPSLibraryBridge {
 
     /**
      * Returns the singleton instance of this library adapter.
+     * <p>
+     * An optional TSS_LIB_WRAPS_SWAP_FILE environment variable may be defined to point to a file name
+     * that will be used as a memory-on-disk for the WRAPS 2.0 native code.
+     * The size of the file is hard-coded in the native code as a static constant due to Rust language specifics.
+     * See src/main/rust/wraps/src/alloc.rs for the definitions.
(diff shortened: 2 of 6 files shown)


rsinha:
Just a side note — the memory usage is independent of the size of the address book.

anthony-swirldslabs:
> Just a side note — the memory usage is independent of the size of the address book.

Thanks for confirming this, @rsinha ! In that case, this solution should work perfectly for any use-cases, including large networks.

rsinha:
LGTM. Great idea! The allocator strategy makes sense, though I haven't verified low-level details like possible off-by-one etc.

rsinha (approved):

andrewb1269 (approved):
Approve the comment in `cryptography/hedera-cryptography-wraps/build.gradle.kts`

Issue #467, opened 2025-11-20
Error handling in WRAPS


rsinha:
We need to avoid the use unwraps / asserts / panics.

rsinha:
https://github.com/hashgraph/hedera-cryptography/pull/565

PR #472, opened 2025-12-06
feat: WRAPS ceremony

rsinha:
**Description**:
This PR introduces necessary definitions and functions for implementing the WRAPS ceremony. There is also a small test using MiMC hashing, which shows an end-to-end test with ceremony, followed by proving and verification.

PS: the unwraps will never be triggered because we only use this code with TSS circuit. That said, there will be a separate PR that does better error handling across all of WRAPS, in response to [this issue](https://github.com/hashgraph/hedera-cryptography/issues/467).


cryptography/hedera-cryptography-wraps/src/main/rust/wraps/src/utils.rs
```diff
@@ -64,4 +68,160 @@ pub fn get_verification_key() -> Result<VerificationKey, WRAPSError> {
...
+/************************************************************************************/
+
+/// Utility type for FFT-related operations on elliptic curve points.
+pub struct ECFFTUtils;
+
+impl ECFFTUtils {
+    /// Computes the FFT of the given data using the Cooley-Tukey algorithm.
+    pub fn fft<C: CurveGroup>(data: &[C::Affine]) -> Vec<C::Affine> {
+        let n = data.len();
+        assert!(n.is_power_of_two());
```
anthony-swirldslabs, line 84: Minor: please consider yielding a Result::Err instead. This can be done separately, as a part of the better unwraps or a similar task.

cryptography/hedera-cryptography-wraps/src/main/rust/wraps/src/utils.rs
```diff
@@ -64,4 +68,160 @@ pub fn get_verification_key() -> Result<VerificationKey, WRAPSError> {
...
+
+
+/************************************************************************************/
+// Methods for FFT-related operations over scalars.
+/************************************************************************************/
+
+/// Utility type for FFT-related operations bound to a specific pairing engine.
+pub struct FFTUtils;
+
+impl FFTUtils {
```
anthony-swirldslabs: Minor: The code here looks somewhat similar to the `ECFFTUtils` above. It would be nice to have a single generics-based solution, so that we don't duplicate it.

cryptography/hedera-cryptography-wraps/src/main/rust/wraps/src/preprocessing.rs
```diff
@@ -0,0 +1,635 @@
...
+    h_g1: Vec<G1Affine>,
+}
+
+pub struct WRAPSPreprocessing {}
+
+impl WRAPSPreprocessing {
+
+    /// Performs the trusted setup for WRAPS, producing the WRAPSProvingKey and WRAPSVerificationKey.
+    pub fn trusted_wraps_setup() -> Result<(WRAPSProvingKey, WRAPSVerificationKey), WRAPSError> {
+        let mut rng = ark_std::rand::rngs::OsRng;
```
anthony-swirldslabs, line 80: Minor q: do we want to provide a seed for deterministic results, or we don't do that on purpose?

cryptography/hedera-cryptography-wraps/src/main/rust/wraps/src/preprocessing.rs
```diff
@@ -0,0 +1,635 @@
...
+        let nova_preprocess_params = PreprocessorParam::new(poseidon_config, F_circuit);
+        // Generate Nova parameters for the WRAPS folding circuit.
+        let (nova_pp, nova_vp) = N::preprocess(
+            &mut rng,
+            &nova_preprocess_params
+        ).map_err(|_| WRAPSError::CryptographyError)?;
+
+        let pp_hash = nova_vp.pp_hash().map_err(|_| WRAPSError::CryptographyError)?;
+
+        let circuit = DeciderEthCircuit::<G1, G2>::dummy((
```
anthony-swirldslabs, line 95: Minor q: "dummy" sounds a bit odd in the context of production code. Is there a "non-dummy" version of that? Or is there a specific reason we use this one here? If so, it would be nice to document that in a comment in code.

cryptography/hedera-cryptography-wraps/src/main/rust/wraps/src/preprocessing.rs
```diff
@@ -0,0 +1,635 @@
...
+        -> Result<(Groth16ProvingKey<PairingCurve>, Groth16VerifyingKey<PairingCurve>), WRAPSError> {
+        let mut rng = ark_std::rand::rngs::OsRng;
+        let (g16_pk, g16_vk) = Groth16::<PairingCurve>::circuit_specific_setup(circuit, &mut rng)
+            .map_err(|e| Error::SNARKSetupFail(e.to_string()))
+            .map_err(|_| WRAPSError::CryptographyError)?;
+
+        Ok((g16_pk, g16_vk))
+    }
+
+    fn dummy_ceremony_groth_setup<C: ConstraintSynthesizer<Fr>>(circuit: C)
```
anthony-swirldslabs: Minor: same comment about the "dummy" part. Perhaps this is on purpose, or maybe we could find a better term for this ("genesis"? - just a random guess though.)

cryptography/hedera-cryptography-wraps/src/main/rust/wraps/src/preprocessing.rs
```diff
@@ -0,0 +1,635 @@
...
+            let num_constraints = cs.num_constraints();
+            abc_g1[start..end].copy_from_slice(&ifft_of_powers_of_beta_tau_g1[(start + num_constraints)..(end + num_constraints)]);
+            a_g1[start..end].copy_from_slice(&ifft_of_powers_of_tau_g1[(start + num_constraints)..(end + num_constraints)]);
+        }
+
+        for (i, u_i) in ifft_of_powers_of_beta_tau_g1.iter().enumerate().take(cs.num_constraints()) {
+            for &(ref coeff, index) in &matrices[0][i] {
+                abc_g1[index] = (abc_g1[index] + (*u_i * coeff)).into_affine();
+            }
+        }
```
anthony-swirldslabs, line 248: Minor: all the math in nested loops in this method, like this one here, could probably run on a GPU with better performance than on a CPU. (I'm assuming the arrays and matrices are pretty large here. I'm also assuming the code deals with floating point or integer numbers here.)

In Java, I used a library called `aparapi` that allowed me to write a kernel in Java and run it on a GPU or automatically fall-back to CPU if a GPU is unavailable. I'm unsure if a similar technology exists for Rust, but if it does, it might be reasonable to consider using it here.

anthony-swirldslabs (dismissed):
Looks good to me. I don't have a deep background in the math involved here, nor am I familiar with the ark libraries, but from reading the code, I didn't find anything obviously wrong. Just a few minor comments.

cryptography/hedera-cryptography-wraps/src/main/rust/wraps/src/utils.rs
rsinha: Yes, I will create a separate PR that does better error handling everywhere.

cryptography/hedera-cryptography-wraps/src/main/rust/wraps/src/utils.rs
rsinha: Yes, I tried, but it was difficult without changing arkworks. 
That said, I realize I dont need the plain FFT -- just the ECFFT. 
I will get rid of it in a cleanup PR, that does other performance improvements.

cryptography/hedera-cryptography-wraps/src/main/rust/wraps/src/preprocessing.rs
rsinha: We could. But we won't have the platform call into this API. It will be purely for ceremony.

cryptography/hedera-cryptography-wraps/src/main/rust/wraps/src/preprocessing.rs
rsinha: Yes, indeed. I am planning to create a separate PR for parallelism, which will come after a PR that creates a ceremony-specific binary artifact (since I want to be able to try a few things and measure performance).

cryptography/hedera-cryptography-wraps/src/main/rust/wraps/src/preprocessing.rs
rsinha: Its a sample ceremony execution, which belongs in a test module. Thanks for the catch.

cryptography/hedera-cryptography-wraps/src/main/rust/wraps/src/utils.rs
rsinha: removed it

cryptography/hedera-cryptography-wraps/src/main/rust/wraps/src/preprocessing.rs
rsinha: This dummy is for a specific reason. We are generating parameters based on the circuit structure, and not a specific instance. So, it is typically called a dummy instance, which has unassigned wire values.

anthony-swirldslabs (approved):

cryptography/hedera-cryptography-wraps/src/main/rust/wraps/src/preprocessing.rs
anthony-swirldslabs: Sounds good.

PR #475, opened 2025-12-10
perf: fine-tune TSS swap

### LINKED ISSUE #474: Fine-tune TSS swap
Come up with TSS swap parameters that reduce RAM usage.

anthony-swirldslabs:
**Description**:
1. Optimizing our custom memory allocator in Rust to be performant with smaller swap block/threshold sizes. And also hardening it by avoiding unsigned integer math overflows.
2. Testing various combinations of parameters and adding a table with the results.
3. Setting the swap size to 24 GB and the threshold/block size to 1 KB.

With the current settings, when the swap is enabled, the library should use only 1.73 GB of RAM (let's call it 2 GB), with the rest of the memory being allocated in our custom swap file, which requires 24 GB of disk space.

If/when our RAM/disk requirements change, we can refer to the added table to fine-tune the parameters further.

**Related issue(s)**:

Fixes #474

**Notes for reviewer**:
All tests should pass. The fix only modifies the code path that runs when the TSS swap is enabled. Locally, with the swap enabled, all tests pass too.

**Checklist**

### DIFF
--- a/cryptography/hedera-cryptography-wraps/src/main/rust/wraps/src/alloc.rs
+++ b/cryptography/hedera-cryptography-wraps/src/main/rust/wraps/src/alloc.rs
@@ -13,11 +13,39 @@ use memmap2::{MmapMut /*, UncheckedAdvice*/};
 use filemap::FileMap;
 use bitmap::BitMap;
 
-/// Allocate small objects in RAM, larger objects on disk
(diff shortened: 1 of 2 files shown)


rsinha (approved):
Looks good, except I havent thought through correctness in all the edge cases etc. -- but, it makes sense.

PR #480, opened 2025-12-18
chore: updating verification api

rsinha:
**Description**:
This PR adds missing checks related to the ledgerID and hinTS verification key in the WRAPS verification method. Now, the caller should be able to verify that the proof is valid with respect to the expected ledgerID (technically, the genesis address book hash).

This PR also renames occurrences of `tss_vk` with `hints_vk` to make it clearer how to use the API.


anthony-swirldslabs (approved):
`lib.rs` changes look good to me. Thanks!

PR #486, opened 2026-01-10
feat: TSS.verifyTSS()

### LINKED ISSUE #482: verifyTSS()
Create a convenience API:

```java
boolean verifyTSS(byte[] ledger_id, byte[] tss_signature, byte[] block_root_hash);
```

where
* ledgerID is Hash(genesis_ab_hash) || Hash(wraps_verification_key)
* tss_signature is wraps_proof || hints_verification_key || hints_signature
* block_root_hash is a "message" in tss-lib terminology

NOTE: consider moving `wraps_proof` to the end, as it may be an aggregate Schnorr signature. The length would tell which one it is. Note that we'll need an extra stateful API to define the Schnorr public keys for this API to work.

Consider adding an API for creating the concatenated arrays as well.

anthony-swirldslabs:
**Description**:
Introducing a convenience API `TSS.verifyTSS(ledgerId, tssSignature, message)` to simplify verifying blocks. Please see the JavaDocs for the documentation.

Also, updating the doc and tests for the dummy hints verification keys to require 1288 zero bytes instead of 1280 as the old value was a mistake.

**Related issue(s)**:

Fixes #482

**Notes for reviewer**:
New (and existing) unit tests should all pass.

**Checklist**


cryptography/hedera-cryptography-wraps/src/main/rust/wraps/src/lib.rs
```diff
@@ -388,7 +388,7 @@ fn pad_addressbook(ab: &AddressBook) -> AddressBook {
 }
 
 /// Hashes the serialized TSS verification key using SHA-256 followed by Poseidon.
```
jsync-swirlds: This seems slightly off.  If I am reading the code correctly, this just calculates the Poseidon hash; not a Poseidon hash of the SHA2-256 hash...

jsync-swirlds (dismissed):
Looks good, one minor nitpick on a comment that might not be accurate.

cryptography/hedera-cryptography-wraps/src/main/java/com/hedera/cryptography/tss/TSS.java
```diff
@@ -0,0 +1,159 @@
...
+     * the `HintsLibraryBridge.aggregateSignatures()` that produced the above `hints_signature`.
+     * In Hiero networks, the `message` is likely a "block_root_hash".
+     * <p>
+     * The `ledgerId` is generally verified via the provided WRAPS compressed proof using a WRAPS verification key
+     * currently installed in the library. During a network genesis, the WRAPS proof may be unavailable for the first
+     * few blocks, in which case the WRAPS proof in the `tssSignature` is replaced with an `aggregate_schnorr_signature`
+     * that has a different length. When this is the case, this method will check if the `aggregate_schnorr_signature`
+     * is valid for the `ledgerId` as a message that was signed, using the Schnorr public keys set via a call to the
+     * `TSS.setSchnorrPublicKeys()` method. If the client didn't set any keys before calling `TSS.verifyTSS(), then
+     * the verification will fail and return false. While it's okay not to set the keys if the client code is certain
```
edward-swirldslabs: small nit on documentation.  If the keys are not set, it looks like exceptions are thrown? (Final else statement.)

cryptography/hedera-cryptography-wraps/src/main/java/com/hedera/cryptography/tss/TSS.java
```diff
@@ -0,0 +1,159 @@
...
+            if (!WRAPS.verifyCompressedProof(abProof, ledgerId, hintsVerificationKey)) {
+                return false;
+            }
+        } else if (abProof.length == AGGREGATE_SCHNORR_SIGNATURE_LENGTH) {
+            if (TSS.schnorrPublicKeys == null) {
+                throw new IllegalStateException("Schnorr public keys haven't been provided");
+            }
+
+            // We sign a rotation message with Schnorr signatures, so we have to prepare it first:
+            final byte[] hintsKeyHash = WRAPS.hashArray(hintsVerificationKey);
```
edward-swirldslabs, line 138: WRAPS.hashArray() may return null.  Is that problematic here?

cryptography/hedera-cryptography-wraps/src/main/rust/wraps/src/lib.rs
anthony-swirldslabs: Nice catch. I fixed the comment.

jsync-swirlds (approved):

rsinha (approved):

andrewb1269 (approved):
Approve `cryptography/hedera-cryptography-wraps/build.gradle.kts`

cryptography/hedera-cryptography-wraps/src/main/java/com/hedera/cryptography/tss/TSS.java
```diff
@@ -0,0 +1,159 @@
...
+            final byte[] rotationMessage = Arrays.copyOf(ledgerId, ledgerId.length + hintsKeyHash.length);
+            System.arraycopy(hintsKeyHash, 0, rotationMessage, ledgerId.length, hintsKeyHash.length);
+
+            if (!WRAPS.verifySignature(TSS.schnorrPublicKeys, rotationMessage, abProof)) {
```
edward-swirldslabs, line 142: Does this call have the behavior that `verifySignature` return true if and only if `TSS.schorrPublicKeys` is a descendant of the ledger id address book?

cryptography/hedera-cryptography-wraps/src/main/java/com/hedera/cryptography/tss/TSS.java
anthony-swirldslabs: I fixed this in the last commit in this PR: [moar comment fix](https://github.com/hashgraph/hedera-cryptography/pull/486/commits/970597690399614e465f5cf983c7f15d4a2f1e06)

So the doc is correct currently.

cryptography/hedera-cryptography-wraps/src/main/java/com/hedera/cryptography/tss/TSS.java
anthony-swirldslabs: In theory, it could be problematic, as in, the convenience API could throw an NPE at line 140 right below.

However, in practice the method would only return `null` if the input array is itself null or empty. Neither of these are possible because we verify that the `tssSignature` array has enough bytes to contain the `hintsVerificationKey` in its head. So the probability of getting a `null` here is practically close to zero. But I can add an extra condition here to make sure the API returns an IllegalArgumentException instead, implying that the provided `tssSignature` was "malformed" somehow.

cryptography/hedera-cryptography-wraps/src/main/java/com/hedera/cryptography/tss/TSS.java
anthony-swirldslabs: I'll let @rsinha confirm, but my understanding is that the `WRAPS.verifySignature()` returns true if the constructed `rotationMessage`, which is a concatenation of the `ledgerId` (aka the genesis AB hash) and the `hintsVerificationKey` (constructed by nodes in the network during a hinTS ceremony), has been signed by the provided `TSS.schnorrPublicKeys` resulting in a given `abProof` (which is an aggregate Schnorr signature in this code path.)

In other words, as of today, this call merely verifies that the given signature is valid for the ledgerId against the provided keys.

Put another way, this validation only works for a Schnorr signature on a rotation message that used the genesis AB hash, aka the `ledgerId` itself. If a future descendent AB hash has been signed with the Schnorr keys, then this specific validation here in this method would fail. In other words, this specific if/else branch DOES NOT support AB rotations with simple Schnorr aggregate signatures. One MUST use WRAPS proofs for AB rotations. This specific branch only support the validation of a few first blocks immediately after the genesis BEFORE any AB rotations take place.

anthony-swirldslabs: In theory, this method could support verifying rotated ABs if one supplies a rotated AB hash instead of the `ledgerId` to this method. However, such a validation would fail the moment WRAPS proofs become available because they need to check against the genesis AB hash, not a rotated AB hash. Also, a validation on a rotated AB hash using the Schnorr signatures would be useless because that signature doesn't maintain a connection with the original genesis AB hash (aka the ledgerId). Only WRAPS proofs support the connection to the ledgerId.

rsinha: @anthony-swirldslabs, your summary is accurate.

cryptography/hedera-cryptography-wraps/src/main/java/com/hedera/cryptography/tss/TSS.java
anthony-swirldslabs: This is fixed at https://github.com/hashgraph/hedera-cryptography/pull/490 .

PR #487, opened 2026-01-12
chore: updating security checks in TSS based on new findings

rsinha:
**Description**:
This PR implements a stronger degree check, which is one of the security checks in hinTS.
It also implements a fix in the checking of hinTS verification key hash in WRAPS.


anthony-swirldslabs (dismissed):

cryptography/hedera-cryptography-wraps/src/main/rust/wraps/src/lib.rs
```diff
@@ -1169,7 +1171,7 @@ mod tests {
             } else {
                 create_new_addressbook()
             };
-            let next_tss_vk = [0u8; 1280]; // placeholder for TSS vk bytes
+            let next_tss_vk = [i as u8; 1480]; // placeholder for TSS vk bytes
```
anthony-swirldslabs, line 1174: Minor: it's unclear why this needs to be initialized with `i` and not zero because it's unclear where the dummy values may possibly be used and how a particular value affects the logic. A bit more elaborate comment in code here could help.

anthony-swirldslabs (approved):

cryptography/hedera-cryptography-wraps/src/main/rust/wraps/src/lib.rs
rsinha: Oh since this is a test of WRAPS, I am only interested in changing the hinTS key, not necessarily using a proper hinTS key. So it is still dummy, but changing.

rsinha: This test failed without the fix.

anthony-swirldslabs: Got it. Didn't realize that was a test code. It looks fine then. Thanks!

Issue #484, opened 2026-01-07
Support replacing WRAPS verification keys


anthony-swirldslabs:
https://github.com/hashgraph/hedera-cryptography/issues/483 hard-coded the WRAPS verification keys in the tss-lib.

This story proposes to introduce an API for client applications to update the keys to support cases when the WRAPS verification keys need an update, and the application needs to verify both old and new WRAPS proofs using different keys.

The API could look something like:

```
public static void setWrapsDeciderVerificationKey(byte[] deciderVK);
```

Also, we need a getter API to retrieve the current key, something like:

```
public static byte[] getWrapsDeciderVerificationKey();
```

Note that the client application would be fully responsible for thread safety when using the new API. This is so by design for performance and API convenience reasons, primarily because the library will have the "default" (or "current") key hard-coded, and we don't want the verification API to accept the key explicitly.

anthony-swirldslabs:
This is being implemented at https://github.com/hashgraph/hedera-cryptography/pull/490

anthony-swirldslabs:
The PR is merged. Resolving.

PR #490, opened 2026-01-13
feat: WRAPSVerificationKey

### LINKED ISSUE #483: Move WRAPS verification key to the library
Today we host the WRAPS keys, both proving and verification, in a 2GB binary artifact at https://builds.hedera.com/tss/hiero/wraps/v0.1/wraps-v0.1.0.tar.gz . A library client is responsible for obtaining the artifact and pointing the library to a directory with the artifact via the `TSS_LIB_WRAPS_ARTIFACTS_PATH` env var.

1. We may need to support changing the keys, although this should be rare.
2. Additionally, we want the verification APIs to be lightweight in general, and a 2GB binary artifact doesn't help here.

So we want to extract the verification keys, the `decider_vp.bin` (1.7KB), and hard-code it as byte a array in the library. This way, the verification API wouldn't depend on the 2GB binary artifact.

In the future, we may consider adding an API to replace the verification key to support its update. See https://github.com/hashgraph/hedera-cryptography/issues/484 .

anthony-swirldslabs:
**Description**:
Introducing `WRAPSVerificationKey` that:
1. Hard-codes the default WRAPS verification key (aka the 1.7 KB `decider_vp.bin` from the 2GB WRAPS binary artifact) in the library and makes the verification API use it.
2. Provides APIs for getting and setting the keys for clients to handle rare scenarios when the key needs to be updated.

Also, addressing a minor comment from another PR at https://github.com/hashgraph/hedera-cryptography/pull/486#discussion_r2682995030

**Related issue(s)**:

Fixes #483 and #484

**Notes for reviewer**:
Existing tests for the bridge and the convenience API call the verification method and therefore test the new feature implicitly. Other than that, the utility class is a trivial setter/getter for a single byte array and doesn't contain any business logic, so there's not any additional tests.

**Checklist**


jsync-swirlds (approved):

rsinha (approved):

xin-hedera (approved):
LGTM

Issue #489, opened 2026-01-13
Add weighted verification API for Schnorr multisigs


rsinha:
This is required for verifying genesis signatures.

PR #491, opened 2026-01-13
feat: schnorr multisig verification w.r.t. addressbook

rsinha:
**Description**:
This PR makes API changes that allow verification of Schnorr aggregate signature w.r.t. to an addressbook, to support genesis case.

Addresses https://github.com/hashgraph/hedera-cryptography/issues/489


anthony-swirldslabs (dismissed):
The `lib.rs` changes look good to me.

Note the Java API changes that I've just pushed:
* `TSS.setSchnorrPublicKeys(keys)` -> `TSS.setAddressBook(keys, weights)`
* `WRAPSLibraryBridge.runSigningProtocolPhase()` accepts weights and signers.
* `WRAPSLibraryBridge.verifySignature()` accepts weights.
* `WRAPSLibraryBridge.constructWrapsProof()` no longer needs signers.

The rest is just plumbing and tests updates.

cryptography/hedera-cryptography-wraps/src/main/java/com/hedera/cryptography/tss/TSS.java
```diff
@@ -13,14 +13,17 @@ public final class TSS {
     private static final int HINTS_VERIFICATION_KEY_LENGTH = 1480;
     private static final int HINTS_SIGNATURE_LENGTH = 1632;
     private static final int COMPRESSED_WRAPS_PROOF_LENGTH = 704;
-    private static final int AGGREGATE_SCHNORR_SIGNATURE_LENGTH = 64;
+    private static final int AGGREGATE_SCHNORR_SIGNATURE_LENGTH = 76;
```
rsinha: @anthony-swirldslabs this should be variable length, depending on the length of the bitvector.

anthony-swirldslabs: I suggest we always pad the bitvector up to the MAX_AB_SIZE, effectively fixing the size of the signature. This solution would be safer than allowing a variable-size signature.

anthony-swirldslabs (approved):
The bitvector padding logic looks good.

PR #494, opened 2026-01-16
chore: remove rpm

### LINKED ISSUE #493: Remove the old rpm module
We replaced the RPM module with the new WRAPS module. We want to remove the old module to reduce the size of the library and eliminate code that we no longer support/maintain.

anthony-swirldslabs:
**Description**:
Removing the no-longer-supported old RPM module from the library so that we reduce its size and optimize its builds in GitHub. The old code will continue to live in the git history of the repository in case we need to refer to it in the future.

The dependency on this old module has been removed from the CN code with https://github.com/hiero-ledger/hiero-consensus-node/pull/22912 .

The PR is 99% removing old files. There's only two material changes in:
* README.md
* aggregation/build.gradle.kts

**Related issue(s)**:

Fixes #493 

**Notes for reviewer**:
All tests should pass.

**Checklist**


andrewb1269 (approved):
Approve gradle and readme changes.

rsinha (approved):

PR #503, opened 2026-01-22
feat: optional cache for ProvingKey

### LINKED ISSUE #500: Cache WRAPS proving key

anthony-swirldslabs:
**Description**:
Introducing a feature to cache the `ProvingKey` in memory instead of re-loading it from disk for every proof construction. The feature is enabled with the `TSS_LIB_WRAPS_ARTIFACTS_CACHE_ENABLED="true"` env var.

Loading and deserializing the key may take ~27 minutes while the construction of the proof itself only takes a few minutes, which is okay when proofs need to be constructed rarely (once a day or once a month - like in mainnet.) However, various tests may benefit from being able to construct proofs quickly, for example to actually test rotating an AddressBook. This feature enables this.

As a part of this fix, I'm enabling the feature when running unit tests via Gradle and unlocking a test that actually performs an AB rotation and verifies the new proof against the genesis AB hash. Previously this extended test was disabled (under `if(true)return`) because it would add an extra 30 minutes for running the tests making the total runtime close to 1 hour. With the cache enabled, the second proof takes almost no time, and all tests run within the same 30 minutes as today. Yet, we now test a lot more.

**Related issue(s)**:

Fixes #500 

**Notes for reviewer**:
All tests, including the newly unlocked one, should pass.

**Checklist**

### DIFF
--- a/cryptography/hedera-cryptography-wraps/src/main/rust/wraps/src/utils.rs
+++ b/cryptography/hedera-cryptography-wraps/src/main/rust/wraps/src/utils.rs
@@ -1,6 +1,7 @@
 // SPDX-License-Identifier: Apache-2.0
 use std::env;
 use std::fs::File;
+use std::sync::OnceLock;
 use memmap2::Mmap;
 use ark_ec::{AffineRepr, CurveGroup};
 use ark_ff::{FftField, Field, Zero};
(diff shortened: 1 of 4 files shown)


cryptography/hedera-cryptography-wraps/src/main/rust/wraps/src/utils.rs
```diff
@@ -30,8 +31,59 @@ pub fn serialize<T: ark_serialize::CanonicalSerialize>(
...
+    /// Create a new cached value wrapper.
+    fn new_ref(ref_pk: &'static ProvingKey) -> Self {
+        ProvingKeyWrapper {
+            owned_pk: None,
+            ref_pk: Some(ref_pk)
+        }
+    }
+
+    /// Return a reference to the value.
+    pub unsafe fn get_ref(&self) -> &ProvingKey {
```
rsinha: Can you please clarify why this need to be `unsafe`?

edward-swirldslabs (dismissed):

cryptography/hedera-cryptography-wraps/src/main/rust/wraps/src/utils.rs
anthony-swirldslabs: It used to unwrap an optional unconditionally, but I've changed the implementation, so this function no longer needs to be `unsafe`. Please see the latest commit where I just fixed this. Thanks for the catch!

rsinha (approved):

edward-swirldslabs (approved):

PR #509, opened 2026-01-27
feat: TSS ceremony logic

rsinha:
**Description**:
This PR implements TSS ceremony protocol. We will have a separate PR creating a CLI that calls the necessary methods in the library. This PR also includes a couple of test cases, one for a smaller circuit (MiMC hashing), and the other being the full TSS circuit -- the full TSS circuit requires a machine with 100s of GBs of RAM, so it is recommended to run the smaller test on smaller machines.


cryptography/hedera-cryptography-wraps/src/main/rust/wraps/src/preprocessing.rs
```diff
@@ -71,6 +78,164 @@ pub struct Phase2SRS {
...
+#[derive(CanonicalDeserialize, CanonicalSerialize)]
+pub struct CircuitConfig {
+    pub num_constraints: usize,
+    pub num_witness_variables: usize,
+    pub num_instance_variables: usize,
+}
+
+fn load_from_file<T: CanonicalDeserialize>(path: &PathBuf) -> Result<T, Error> {
+    let raw_data = std::fs::read(path)?;
+    let data: T = T::deserialize_uncompressed(&*raw_data)?;
```
anthony-swirldslabs, line 90: Minor: doesn't ark-serialize io provide a `Read` implementation that can read directly from a file? This could help reduce the RAM usage for large files.

cryptography/hedera-cryptography-wraps/src/main/rust/wraps/src/preprocessing.rs
```diff
@@ -71,6 +78,164 @@ pub struct Phase2SRS {
...
+    Ok(data)
+}
+
+fn store_to_file<T: CanonicalSerialize>(path: &PathBuf, data: &T) -> Result<(), Error> {
+    let mut raw_data = Vec::new();
+    data.serialize_uncompressed(&mut raw_data)?;
+    std::fs::write(path, &raw_data)?;
```
anthony-swirldslabs, line 97: Minor: similar to the above, is there a `Write` implementation that could write directly to a file?

cryptography/hedera-cryptography-wraps/src/main/rust/wraps/src/preprocessing.rs
```diff
@@ -71,6 +78,164 @@ pub struct Phase2SRS {
...
+const PAUSE: bool = false;
+fn pause_until_enter() {
+    if PAUSE {
+        print!("Press Enter to continue...");
+        std::io::stdout().flush().unwrap(); // make sure prompt shows before blocking
+
+        let mut s = String::new();
+        std::io::stdin().read_line(&mut s).unwrap(); // waits for Enter
+    }
+}
```
anthony-swirldslabs, line 110: Minor: this logic might belong to the CLI tool implementation rather than be a part of this library code.

cryptography/hedera-cryptography-wraps/src/main/rust/wraps/src/preprocessing.rs
```diff
@@ -71,6 +78,164 @@ pub struct Phase2SRS {
...
+            .enumerate()
+            .map(|(i, g)| {
+                let tau_pow_i = multiplier * tau.pow([i as u64]);
+                g.mul(tau_pow_i).into_affine()
+            })
+            .collect::<Vec<G1Affine>>();
+        store_to_file::<Vec<G1Affine>>(&next_srs.join(name), &new_powers_of_tau_g1).unwrap();
+        drop(new_powers_of_tau_g1);
+        drop(prev_powers_of_tau_g1);
+    } else {
```
anthony-swirldslabs, line 126: Minor q: why are the explicit `drop()` calls needed? Aren't they supposed to be called implicitly at line 126 on the `}` as they go out of scope? Same q for a few methods below.

cryptography/hedera-cryptography-wraps/src/main/rust/wraps/src/preprocessing.rs
```diff
@@ -130,34 +295,48 @@ impl WRAPSPreprocessing {
...
         let domain = GeneralEvaluationDomain::<Fr>::new(n)
             .ok_or(SynthesisError::PolynomialDegreeTooLarge).unwrap();
         let domain_size = domain.size();
 
+        // generate default SRS elements and store them to disk, so we can purge them from memory
+
+        // generate default powers of tau in G1
         let powers_of_tau_g1: Vec<G1Affine> = vec![G1Affine::generator(); 2*domain_size]; // {Gx:i} | i=0..2n-2}
+        store_to_file::<Vec<G1Affine>>(&output_path.join("powers_of_tau_g1.bin"), &powers_of_tau_g1).unwrap();
+        drop(powers_of_tau_g1);
```
anthony-swirldslabs, line 312: Minor: I totally realize why you're calling drop() in this method. So my comment above doesn't apply here. It only applies to those few methods above that call it immediately before a closing `}`.

cryptography/hedera-cryptography-wraps/src/main/rust/wraps/src/preprocessing.rs
```diff
@@ -166,185 +345,210 @@ impl WRAPSPreprocessing {
...
+        let start = std::time::Instant::now();
+        specialize_srs_helper_g1(&mut abc_g1, &matrix_b_path, &ifft_of_powers_of_alpha_tau_g1, num_constraints, num_variables);
+        println!("Updating abc_g1 using powers_of_alpha_tau_g1 took {:?}", start.elapsed());
 
-        for (i, u_i) in ifft_of_powers_of_alpha_tau_g1.iter().enumerate().take(cs.num_constraints()) {
-            for &(ref coeff, index) in &matrices[1][i] {
-                abc_g1[index] = (abc_g1[index] + (*u_i * coeff)).into_affine();
-            }
-        }
+        pause_until_enter();
```
anthony-swirldslabs, line 486: Minor: to add to my above comment, this looks like a manual debugging break point for local testing, rather than something that's supposed to be used in a production environment.

Does this code actually overwrites any files multiple times to justify having this breakpoint here? If not, then one could examine the already written files while the code keeps running and let the code keep producing new files (like I said, as long as the code doesn't overwrite old files - which it probably shouldn't anyway.)

anthony-swirldslabs (approved):
LGTM (disclaimer: I barely (read - not at all) understand the math happening here, but the code itself does look good to me.)

I left a few minor comments, please feel free to address them in a follow-up PR if appropriate.

PR #514, opened 2026-02-04
feat: introduce TSS Ceremony Orchestrator

### LINKED ISSUE #506: TSS Ceremony Orchestrator
Need to develop a script/app that would implement the orchestration per:

https://www.notion.so/swirldslabs/Basic-Processing-Approach-2be7c9ab25918007bdafc900bf966cae

using a CLI Rust binary that crunches the data.

anthony-swirldslabs:
**Description**:
Introducing an implementation for the TSS Ceremony Orchestrator.
1. Please focus on the Gradle, `scripts/`, `src/test`, and `src/main/java/com/hedera/cryptography/ceremony/*` (note, not the double *) files that actually implement it
2. The `ceremony/crypto/*` and `ceremony/s3/*` are slightly refactored copies of the Consensus Node key loader and Block Node S3 client respectively. I didn't write this code myself. If you have comments on this code, then it makes sense to open issues against the corresponding source projects. It would be unreasonable to introduce modifications to these implementations in our tss-lib fork.

@jsync-swirlds : Please review the whole thing.
@edward-swirldslabs : Please review the Crypto.java/Test thing.
@rsinha : You may be most interested in the DataCruncher.java/Test code to ensure it aligns with your implementation w.r.t. command-line arguments.
But any and all comments are welcome regarding anything you spot.

The Orchestrator is a CLI utility that can be invoked like:
```shell
$ TSS_CEREMONY_S3_ACCESS_KEY=a TSS_CEREMONY_S3_SECRET_KEY=b java -cp ./bcprov-jdk18on-1.81.jar:./bcpkix-jdk18on-1.81.jar:./bcutil-jdk18on-1.81.jar:build/libs/hedera-cryptography-ceremony-3.5.0.jar com.hedera.cryptography.ceremony.Orchestrator 1 1,2, /bin/false us-east-1 "https://s3.amazonaws.com/" TSSCeremonyBucket build/resources/test/ password
```

NOTE: I have to add the Bouncy Castle JARs to the class path manually. So far I didn't find an easy way to create a fat jar or something similar to simplify deploying and running the orchestrator. I suggest we resolve the deployment issues in a separate PR in the future as it may add an extra complexity in the Gradle setup which I'd prefer to keep outside of this main implementation PR.

However, the newly created code is covered with unit tests, including all the utility and the main Phase implementation class. Also, the above command line technically works (as in, it starts the orchestrator alright, but that won't run for too long w/o proper S3 credentials.) Once we resolve the deployment issues described above, we can start testing this code in a real environment to see how it works. For now, the unit tests provide a sufficient level of confidence that the code works per the design.

**Related issue(s)**:

Fixes #506

**Notes for reviewer**:
All tests should pass.

**Checklist**


cryptography/hedera-cryptography-ceremony/src/main/java/com/hedera/cryptography/ceremony/S3DirectoryAccessor.java
```diff
@@ -0,0 +1,124 @@
...
+    }
+
+    /// Download all files from bucket/dir/directoryName/* in S3 to a temp directory in local file system.
+    Path downloadDir(String directoryName) throws IOException {
+        final Path path = Files.createTempDirectory(directoryName);
+
+        try {
+            final String prefix = dir + ensureSlash(directoryName);
+
+            final List<String> objects = s3Client.listObjects(prefix, MAX_OBJECTS);
```
edward-swirldslabs, line 82: Would we ever get close to the max number of objects?

cryptography/hedera-cryptography-ceremony/src/main/java/com/hedera/cryptography/ceremony/Phase.java
```diff
@@ -0,0 +1,142 @@
...
+            inputDir = s3DirectoryAccessor.downloadDir(inputFileName + ".bin");
+
+            final String outputDirName = thisFileName + ".bin";
+            System.err.println("Creating output dir: " + outputDirName);
+            final Path outputDir;
+            outputDir = Files.createTempDirectory(outputDirName);
+
+            System.err.println("Processing input " + inputDir.toAbsolutePath().toString() + " and output result to "
+                    + outputDir.toAbsolutePath().toString());
+            final int exitCode = dataCruncher.execute(phase, inputDir, outputDir);
```
edward-swirldslabs, line 118: Is there any kind of input verification that is necessary?  Is there any way for the contents of the s3Directory to become corrupted?

edward-swirldslabs (approved):

cryptography/hedera-cryptography-ceremony/src/main/java/com/hedera/cryptography/ceremony/S3DirectoryAccessor.java
anthony-swirldslabs: `MAX_OBJECT = 1000`. However, we also sign every file (and add a certificate.pem as well). So the maximum number of files that the dataCruncher should produce is equal to (1000-1)/2 = 499. If we ever go above this limit, then we'll need to make changes in the S3Client implementation to be able to handle more files because that's where this limit is coming from.

@rsinha : Could you please confirm if the number of files that the dataCruncher produces for a single phase is below **499** ? (Please remember that we want each file to have a size less than 5GB, so that we don't need to implement a multi-part downloading in the S3Client.)

cryptography/hedera-cryptography-ceremony/src/main/java/com/hedera/cryptography/ceremony/Phase.java
anthony-swirldslabs: I don't believe the Orchestrator is supposed to understand the data that is being crunched (because the separation of responsibilities and such), so it cannot reasonably validate the input/output files. If the dataCruncher fails to process the input, then it would exit with a non-zero code which we handle here (it can even crash and we'd still be okay because it's running as a separate process.) A failure to upload all the produced files would also be caught as an exception thrown from the uploadDir() call below. So we'd see this in the log (which we push to stderr, and devops will redirect into a file for us.)

Once in S3, I'd guess the only way for the data to be corrupted is for someone to manually modify it. The Orchestrator implementation would only ever upload files using file and directory names that always include the local nodeId which is an argument passed to the Orchestrator. As long as every node starts the orchestrator using its own unique nodeId, they wouldn't be able (aka shouldn't - barring bugs) to overwrite each others files. From this reasoning, I think that the probability of data becoming corrupted in S3 is neglegible.

cryptography/hedera-cryptography-ceremony/src/main/java/com/hedera/cryptography/ceremony/Orchestrator.java
```diff
@@ -0,0 +1,62 @@
...
+
+            // Run phase 1
+            new Phase("1", thisNodeId, allNodeIds, new S3DirectoryAccessor(s3Client, "phase1"), dataCruncher, crypto)
+                    .run();
+
+            // Run phase 2
+            new Phase("2", thisNodeId, allNodeIds, new S3DirectoryAccessor(s3Client, "phase2"), dataCruncher, crypto)
+                    .run();
+        }
+    }
```
jsync-swirlds, line 61: We might need to repeat the process a couple times if not enough systems process the data in one pass.
It would be good to enclose this data in a `while(true)` loop so it reruns if the data is uploaded a second (or third) time.

anthony-swirldslabs: I understand the concern and the intention of the proposed solution. However, there's numerous edge cases which we need to figure out how to handle.

The crux of the problem is that we've implemented the skipping logic where a node may accept input from another node with a much lower nodeId index. An extreme hypothetical example would be to consider a network [1, 2, 3] where both 1 and 2 could not process anything, so the 3 ended up processing the `initial` data. At this point we have
* initial.bin/*
* initial.ready
* 3.claimed with "initial"
* 3.bin/*
* 3.ready

Suppose we added a `while(true)` and the nodes 1 and 2 "woke up" on the second iteration (just realized, there'll be a problem if they wake up while the 3 is crunching data. anyway...) So on this second iteration, we'll get 1 and 2 to crunch the data and produce the output, and then the 3rd node would repeat the processing, only this time it'd be using the `2.bin/*` instead of the `initial.bin/*` as its input. And this process will keep repeating and overwriting the data over and over. Note that the data is random and non-deterministic. If we capture the `3.bin/*` output at some point and use that as our WRAPS data but the nodes keep running the 2nd, 3rd, and subsequent iterations, then the newly overwritten intermediate results will no longer correspond to the final product that we captured, and hence, even though their signatures could be verified, the data won't really directly relate to the results that we captured in the past.

From this extreme example, it's obvious that we cannot let the ceremony run repeatedly in a while(true) loop unless we add more complexity.

So the question is what complexity we want to add and where?

Would introducing a concept of a "cycle" that would represent a single run work? In S3, the directory structure would be bucket/cycle1/phase1, bucket/cycle1/phase2/, bucket/cycle2/phase1/... etc. So the `while(true)` would first check which cycle has been finished (which is also an interesting problem to solve...), and then would start a new cycle by prefixing the phase folder with the cycle folder name when constructing an S3DirectoryAccessor instance.

Does this complexity sound reasonable? How do you propose to solve the "detect a completed cycle" thing given that we allow nodes not to participate (because of hardware failures, timeouts, and whatnot)?

jsync-swirlds: First, I haven't finished reading everything, so I perhaps assumed more existing checks than are present.

My thought was that each node would check 3 things.
1. Has the previous node completed (or exceeded its deadline to start)
1. Has _this_ node **not started**, and have all _later_, nodes **not started**
   * Is there available input data _more recent_ than any later node that has started.
1. Is this node ready to process data for this phase.

If the node gets a "yes" to all three questions, then it continues, otherwise it waits a bit and then checks for all three questions again.

I think we need all three questions regardless, to avoid unintentionally getting things out of order. Given that it seems that adding looping wouldn't be a lot more complexity.

anthony-swirldslabs: Indeed, it may be useful to read the exact checks and logic implemented in the Phase class to see if they match the expectations and/or are reasonable. From there, we can work out how to resolve the original concern.

However, regarding the second check in your list - it explicitly suggests a node to see if later nodes not started. If I interpret this correctly, it means that if the node 2 sees that the node 3 has already started working, then the 2 will never start processing. Same story with the node 1. In the context of the extreme example that I provided above, this would mean that even if the while(true) is present, then the nodes 1 and 2 would never compute anything because the node 3 has finished the computation on the first iteration (hopefully; it may have failed actually.. another edge case.) Which effectively makes the while(true) completely useless (again, unless extra complexity is added somewhere.) It would only help in an unrealistic hypothetical situation where only the later nodes fail all at once, and so the repeated while(true) iterations would allow them to process the data iteration by iteration. I find such an orderly failure very improbable, and therefore a while(true) w/o any additional extra complexity not helping anything.

jsync-swirlds: I think you might have missed the sub-point on the second question.
If node 2 sees that node 3 started, it cannot/must not start _for this input set_ because node 3 already began.
What node 2 can do, however, is wait for a _newer_ input (i.e. we did not get enough nodes the first time, so we upload the initial data again and try a second time).
that newer input effectively starts everything over and the nodes will run through the whole process because each time they see that later nodes have started, they also see that the input is newer than that node start file, so they can safely process it.

andrewb1269 (approved):
Approve the three gradle files.

cryptography/hedera-cryptography-ceremony/src/main/java/com/hedera/cryptography/ceremony/Orchestrator.java
anthony-swirldslabs: Our S3 client doesn't expose object timestamps, and nodes don't delete their input upon uploading their output (and I suppose we don't want to delete it, otherwise we wouldn't bother signing the output in the first place.) Therefore, we currently don't have a mechanism to detect a "newer" initial input.

As proposed above, the concept of "cycles" could in fact introduce just that.

Another alternative would be to rely on a human operator to nuke all the data in S3 that nodes have created, leaving only the initial.ready/bin files (or also replacing these as well), thus restarting the cycle anew if we have the while(true) in there. The disadvantage of this is that we lose the results of the previous cycle unless we manually back them up (e.g. to still go and use them in case further cycles still fail to gather data from more nodes.)

Do you prefer fully automated cycles, or you prefer to include a manual human action into the process?

jsync-swirlds: I see.

I wonder if what we could do is add a date/time value into the "started" and "ended" files written to S3; then the other steps could read that data and get date/time.
It's not a lot to download, and makes timing management possible.  It doesn't really resolve all the other concerns, however.

Creating cycle folders would work as well, though it's a bit more complexity in the orchestration to determine where to look for data. This would leave more leftover detritus from failed runs (which we'd need to remove later), but if it avoids excessive complexity in the orchestration, that's a reasonable trade.

Automated is strongly preferred, there is already enough manual steps to run the compute-intensive work at the beginning and end of each phase; the more autonomous the orchestration is, the better.

anthony-swirldslabs: Re: automation - that's what I was thinking. Thanks for confirming that we're aligned here.

Adding timestamps to claimed/ready files isn't impossible. However, that doesn't solve the problem of a node rebooting (e.g. it crashed and NodeOps restarted it.) It will have no way to determine whether the initial data is "newer" or not. Therefore, the timestamps would add a tiny bit of complexity in the process, but wouldn't really solve (m)any problems.

I agree that cycles would also add some additional complexity into the implementation. The complexity would be marginal where we just add a layer of folders in S3. But the unresolved issue (see below) might be tough to solve. Anyway, the advantage of them is that they fully automate the process. The Orchestrator simply keeps looping in the while(true) and starts a new cycle whenever it appears in S3, allowing a human operator to follow the exact same procedure for any cycle, and then only delete old cycles data when a decision is made to use the last cycle.

As mentioned above, the only tricky part with cycles is to detect what it means for a cycle to be complete. Per the above comments, it sounds that we do want to add one more condition for nodes to decide if they should process the data - it's to look if later nodes have already started their processing and retreat if so (aka do nothing.) I find this reasonable. This also helps resolve the issue when a node was down/restarted in the middle of the process.

And if this is so, then a human would be able to detect when a cycle is complete - once no new files appear from later nodes, and the last output artifacts timestamp exceeds all reasonable timeouts for the remaining nodes (as you've seen, the `WAIT_FOR_REGULAR_FILE_TIMEOUT_MILLIS` is set to 12 hours currently, so the operator would simply multiply the number of remaining nodes by 12 hours to figure this out. I'm happy to reduce the timeout if desired though.)

This way, the human would be able to solve this tricky part - if a decision is made to run a new cycle, then the human would manually create a cycle2/ (or 3 etc.) in S3 and upload initial.bin/ready files into that cycle folder. The orchestrator would see a fresh cycle folder and would start running this new cycle.

Should I go ahead and implement this? Can we do this in a follow-up PR to leverage the existing approvals and make this extra change easier to review (because the PR size and stuff)?

jsync-swirlds: This sounds good. I agree with your plan.
It can be implemented in a follow-up PR.

We probably should also shrink the base timeout duration from 12 hours to 6, so the full process doesn't take excessively long, that can also be in follow-up(assuming that makes sense to you).

I still need to read through the rest of the code (time this week has been in short supply), but I can gather suggestions for your follow-up separately.

PR #515, opened 2026-02-05
fix: hints verification optimization

### LINKED ISSUE #513: HinTS Verification Optimization
Saving one pairing in the hinTS verification algorithm and at the same time 192 bytes of space from the hinTS verification key.

pratyay85:
Description: This PR optimizes the hinTS verification process by reducing the aggregated verification key size and eliminating a pairing computation.

- Remove one element from the aggregated verification key

- Eliminate one pairing computation from the verification logic

- Update the test suite to separately report the time taken for verification

Related issue(s):
Fixes #513

Notes for reviewer: The optimization results in the following performance improvements:

Storage: 192 bytes saved.

Speed: Verification is approximately 1.5 msec faster.

Logs: The test output now explicitly prints the standalone verification time.

Checklist

[x] Documented (Code comments, README, etc.)

[x] Tested (unit, integration, etc.)

### DIFF
--- a/cryptography/hedera-cryptography-hints/src/main/rust/jni_crs.rs
+++ b/cryptography/hedera-cryptography-hints/src/main/rust/jni_crs.rs
@@ -1,11 +1,11 @@
 // SPDX-License-Identifier: Apache-2.0
 
-use jni::objects::{JByteArray, JObject};
-use jni::sys::{jbyteArray, jboolean, jshort};
-use jni::JNIEnv;
-use crate::setup::{ContributionProof, PowersOfTauProtocol};
 use crate::hints::{serialize, CRS};
 use crate::jni_util;
+use crate::setup::{ContributionProof, PowersOfTauProtocol};
+use jni::objects::{JByteArray, JObject};
+use jni::sys::{jboolean, jbyteArray, jshort};
+use jni::JNIEnv;
 
 /// JNI function to create a new CRS object
 /// # Arguments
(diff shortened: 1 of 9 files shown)


cryptography/hedera-cryptography-hints/src/main/rust/jni_crs.rs
```diff
@@ -42,28 +42,28 @@ pub extern "system" fn Java_com_hedera_cryptography_hints_HintsLibraryBridge_upd
     prev_crs_array: JByteArray,
     random_array: JByteArray,
 ) -> jbyteArray {
-    let prev_crs :CRS = match jni_util::deserialize_jbyte_array(&env, &prev_crs_array) {
+    let prev_crs: CRS = match jni_util::deserialize_jbyte_array(&env, &prev_crs_array) {
         Ok(val) => val,
-        Err(_) => return std::ptr::null_mut()
+        Err(_) => return std::ptr::null_mut(),
```
anthony-swirldslabs, line 47: I suggest to revert any and all code reformatting in this PR. Different IDEs of different developers may want to auto-reformat code in its own locally-configured style, causing each PR/commit from different developers to keep reformatting the same code back and forth. This is unproductive because it makes the git history extremely unusable, and it unreasonably increases PR sizes with changes that are unrelated to the core of a fix.

If you want to install a code-formatting standard in this repository, then this would be best achieved using build-time Gradle and/or Cargo checks that would verify the formatting and provide an option to reformat the code per their standards (e.g. `./gradlew spotlessApply` will reformat the Java code today.) This would ensure a consistent code-formatting and avoid any unnecessary massive code reformatting in PRs. If you do want to implement this, then I suggest to do this in a separate dedicated PR to avoid mixing business logic changes with code reformatting changes.

PS. Minor: personally, I'm not a fan of trailing commas because I will never guess if you forgot to add the last item to the list, allowing subtle bugs to slip through.

rsinha:
Replaced with new pr.

PR #518, opened 2026-02-07
feat: support TSS Ceremony cycles

### LINKED ISSUE #517: Support multiple TSS Ceremony cycles

anthony-swirldslabs:
**Description**:
Introducing a concept of the TSS Ceremony cycles to the Orchestrator. This allows us to run multiple cycles in case we're not satisfied with the results of the previous runs.

In S3, the directory structure looks like:
* cycle0/phase1/initial.bin/*
* cycle0/phase1/initial.ready
* ... and the above starts the phase1 of cycle0, nodes populate S3 with data, and run phase1 and then phase2.

Once complete, the Orchestrator will re-run this phase over and over. However, the phase implementation detects existing .claimed files and skips actual data processing. So the data computed at the cycle0 remains intact.

If we aren't satisfied with the results, then the operator (aka @rsinha ) goes and adds the following to S3:
* cycle1/phase1/initial.bin/*
* cycle1/phase1/initial.ready

The Orchestrator will detect this and will run the cycle1.

This can continue up to cycle99 which is the maximum number of supported cycles.

**Related issue(s)**:

Fixes #517

**Notes for reviewer**:
All tests should pass.

**Checklist**

### DIFF
--- a/cryptography/hedera-cryptography-ceremony/src/main/java/com/hedera/cryptography/ceremony/Orchestrator.java
+++ b/cryptography/hedera-cryptography-ceremony/src/main/java/com/hedera/cryptography/ceremony/Orchestrator.java
@@ -3,10 +3,24 @@ package com.hedera.cryptography.ceremony;
 
 import com.hedera.cryptography.ceremony.s3.S3Client;
 import com.hedera.cryptography.ceremony.s3.S3ClientInitializationException;
+import com.hedera.cryptography.ceremony.s3.S3ResponseException;
+import java.io.IOException;
 import java.util.Arrays;
 import java.util.List;
 
+/// TSS Ceremony Orchestrator.
 public class Orchestrator {
+    /// Maximum number of cycles of the ceremony.
+    private static final int MAX_CYCLES = 100;
+
+    /// Make the loop less tight to reduce the number of S3 API calls in case we keep re-running the last cycle.
+    /// Long enough to avoid bombarding S3 with unnecessary requests.
+    /// Short enough to allow a human to start a new cycle by uploading the initial.bin/ready whenever needed.
+    private static final long WAIT_BETWEEN_CYCLES_MILLIS = 15 * 60 * 1000;
+
+    /// Make the loop less tight to avoid 429 TOO_MANY_REQUESTS.
+    private static final long WAIT_IN_DETERMINE_CYCLE_LOOP_MILLIS = 1000;
+
     /**
      * Main class that requires command-line arguments:
      * <p>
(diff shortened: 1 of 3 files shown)


cryptography/hedera-cryptography-ceremony/src/main/java/com/hedera/cryptography/ceremony/Orchestrator.java
```diff
@@ -50,13 +64,79 @@ public static void main(String[] args) throws S3ClientInitializationException {
...
+
+    private static String cycleName(int i) {
+        return "cycle" + i;
+    }
+
+    /// Determine a cycle to run, or null if there's none ready.
+    private static String determineCycle(S3Client s3Client) throws S3ResponseException, IOException {
+        // It would be nice to list all objects and filter out the pattern we like, but S3 API has a hard-limit
+        // of 1000 max_keys for listObject call, and with multiple cycles and phases we could hit the limit.
+        // So we loop instead:
```
jsync-swirlds, line 124: Could we use the [Bucky](https://github.com/hashgraph/bucky) client instead of a local copy?  If so would that work around this limit?

jsync-swirlds (approved):

cryptography/hedera-cryptography-ceremony/src/main/java/com/hedera/cryptography/ceremony/Orchestrator.java
anthony-swirldslabs: We couldn't, for a few reasons:
1. You may have access to that repo. I presume Jasper and a few other folks too. Myself, I get 404, and I guess everyone else gets a 404. From this, I presume that this closed repo isn't widely used in production, and hence I don't trust its implementation as much as I trust the code found in the open BN repo.
2. This limitation comes from the S3 API specification rather than from our particular client implementation. Yes, the S3 API provides for pagination. However, our S3 client doesn't support it. It may not be the worst idea to add the support, but that's an extra complexity and hence an extra source of bugs.
3. From the perspective of the amount of data that our code needs to process here, it may be that the chosen approach isn't too bad. E.g., if we had a couple of cycles, each with 2 phases, each generating (random guess) 10 files, we'd be dealing with 2 * 2 * 10 * 2 = 80 objects + initial files + claimed/ready markers ~~ 100 objects I'd say. And this particular code doesn't really need to know anything about 99.9% of them. It only really cares about the initial.ready files and nothing else. So I'm guessing that the current implementation is good for our use-case.

jsync-swirlds: 1. Bucky is a straight copy from BN, with a few bug fixes; BN will switch to Bucky in the current sprint.
1. Understood; I think the Bucky client does support the pagination, but I could be incorrect.
1. This is reasonable; just thinking of ways to minimize custom code where it makes sense.

rsinha (approved):

cryptography/hedera-cryptography-ceremony/src/main/java/com/hedera/cryptography/ceremony/Orchestrator.java
anthony-swirldslabs: 1. You want to open the repo for that I suppose.
2. Maybe. But the code in BN, specifically the code in the `listObjects` does not. I cannot confirm if bucky supports it, because like I said I cannot open the repo.
3. +1. In fact my math was assuming we only have 1 node. If we had 40 nodes, then the num of objects would explode to 2 * 2 * 10 * 40 * 2 = 3200, which is a lot. And this code still only cares about 2 files only - the cycle0/phase1/initial.ready and cycle1/phase1/initial.ready in this particular setup. So yes, I believe the current implementation is the best we can have.

PR #522, opened 2026-02-10
feat: ceremony cli

rsinha:
**Description**:
This PR implements the ceremony CLI, that can be invoked by the orchestrator.

### DIFF
--- a/cryptography/hedera-cryptography-wraps/src/main/rust/ceremony/Cargo.toml
+++ b/cryptography/hedera-cryptography-wraps/src/main/rust/ceremony/Cargo.toml
@@ -1,19 +1,17 @@
 [package]
-name = "wraps"
+name = "ceremony"
 version = "0.1.0"
 edition.workspace = true
 
-[lib]
-name = "wraps"
-path = "src/lib.rs"
-crate-type = ["cdylib", "rlib"]
 
 [dependencies]
+clap = { version = "4", features = ["derive"] }
 ark-ec = { workspace = true }
 ark-ff = { workspace = true, features = ["asm"] }
 ark-poly = { workspace = true }
 ark-std = { workspace = true, features = ["getrandom"] }
 ark-crypto-primitives = { workspace = true, features = ["constraints", "sponge", "crh", "prf"] }
+ark-poly-commit = { workspace = true }
 ark-relations = { workspace = true }
 ark-r1cs-std = { workspace = true }
 ark-snark = { workspace = true }
@@ -25,12 +23,11 @@ ark-grumpkin = { workspace = true, features = ["r1cs"] }
 jni = "0.21.1"
 memmap2 = "0.9.9"
 rayon = { version = "1.10.0" }
+sha3 = { workspace = true }
 blake2 = { workspace = true }
 digest = { workspace = true }
-rand = { workspace = true }
-rand_chacha = { workspace = true }
 
-folding-schemes = { workspace = true }
+wraps = { path = "../wraps" }
 
 [profile.dev]
 opt-level = 3
--- a/cryptography/hedera-cryptography-wraps/src/main/rust/ceremony/Cargo.toml
+++ b/cryptography/hedera-cryptography-wraps/src/main/rust/ceremony/Cargo.toml
@@ -31,3 +31,6 @@ wraps = { path = "../wraps" }
 
 [profile.dev]
 opt-level = 3
+
+[profile.release]
+opt-level = 3
--- a/cryptography/hedera-cryptography-wraps/src/main/rust/ceremony/src/main.rs
+++ b/cryptography/hedera-cryptography-wraps/src/main/rust/ceremony/src/main.rs
@@ -9,58 +9,45 @@ use wraps::preprocessing::WRAPSPreprocessing;
     about = "Run WRAPS Groth16 ceremony preprocessing phases",
     long_about = "Run WRAPS Groth16 ceremony preprocessing phases.\n\
 \n\
-Required base args:\n\
-  --phase <0..5>\n\
-  --output-folder <PATH> (used in phases 1, 2, 4, 5)\n\
+Positional CLI format:\n\
+  ceremony <phase> [arg1] [arg2] [arg3] [arg4]\n\
 \n\
-Folder mapping by phase:\n\
+Positional argument meaning by phase:\n\
   phase 0 (extract_circuit_r1cs_config):\n\
-    circuit-folder= circuit\n\
+    arg1 = circuit-folder\n\
 \n\
   phase 1 (create_init_srs_phase1):\n\
-    circuit-folder= circuit\n\
-    output= phase1_init\n\
+    arg1 = circuit-folder\n\
+    arg2 = output-folder\n\
 \n\
   phase 2 (update_srs_phase1):\n\
-    circuit-folder= circuit\n\
-    input-folder= prev_phase1_srs\n\
-    output= next_phase1_srs\n\
+    arg1 = circuit-folder\n\
+    arg2 = input-folder\n\
+    arg3 = output-folder\n\
 \n\
   phase 3 (specialize_srs):\n\
-    circuit-folder= circuit\n\
-    phase1-input-folder= final_phase1_srs\n\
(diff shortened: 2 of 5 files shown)


cryptography/hedera-cryptography-wraps/src/main/rust/ceremony/src/main.rs
```diff
@@ -0,0 +1,220 @@
...
+
+#[derive(Debug, Parser)]
+#[command(
+    name = "ceremony",
+    about = "Run WRAPS Groth16 ceremony preprocessing phases",
+    long_about = "Run WRAPS Groth16 ceremony preprocessing phases.\n\
+\n\
+Required base args:\n\
+  --phase <1..5>\n\
+  --output-folder <PATH> (used in phases 1, 2, 4, 5)\n\
```
anthony-swirldslabs: The current implementation of the DataCruncher class in the Orchestrator uses positional arguments rather than named arguments. It would introduce an unnecessary complexity to add name prefixes, especially when they change from phase to phase.

Could you please consider dropping the names (like `--phase` etc.) and simply rely on the position of the argument on the command line?

To provide an example, the DataCruncher class would invoke the executable literally like (using hypothetical paths):
```shell
$ run_ceremony 2 /tmp/params /tmp/input /tmp/output
```

cryptography/hedera-cryptography-wraps/src/main/rust/ceremony/Cargo.toml
```diff
@@ -0,0 +1,36 @@
...
+ark-relations = { workspace = true }
+ark-r1cs-std = { workspace = true }
+ark-snark = { workspace = true }
+ark-serialize = { workspace = true }
+ark-groth16 = { workspace = true }
+ark-bn254 = { workspace = true, features = ["r1cs"] }
+ark-ed-on-bn254 = { workspace = true, features = ["r1cs"] }
+ark-grumpkin = { workspace = true, features = ["r1cs"] }
+jni = "0.21.1"
+memmap2 = "0.9.9"
```
anthony-swirldslabs, line 24: Minor: I'm unsure why these two need to pop up in the ceremony CLI tool. It doesn't interact with Java, nor does it (currently) use memory-mapped files on disk. But I suspect that the existing wraps code may be pulling them.

Further, building-wise, it may be complex to produce a second separate binary in the same Gradle project in our current setup because AFAIK it can only build one by default, see: https://github.com/hashgraph/hedera-cryptography/blob/main/cryptography/hedera-cryptography-wraps/build.gradle.kts#L9
As you recall, we had to introduce quite a bit of complexity in order to build a compressor binary along with the raps library back in the day.

Moreover, it's unclear if the ceremony binary needs to be packaged in the wraps artifacts directly. The artifacts are primarily used to construct a proof (rarely) or validate a proof (often.) The ceremony may need to be run once in a life time (ideally), and adding the binary to the artifacts seems wasteful then because it will increase the size w/o adding any value for daily operations.

It's for this reason that I put the Orchestrator into a separate Gradle project. So we may want to consider extracting (ideally) or copying (less ideally) the ceremony-related code into that separate Gradle module together with the Orchestrator. If/when this is done, then I believe that neither the jni nor the memmap2 dependencies should be required then, because the ceremony code shouldn't really depend on these.

But we can do that later in a separate fix if desired.

PS. Just to clarify the required end result, we need `rm -rf ./build && ./gradlew build` to actually produce the binary for this tool. I don't believe this command will do that in the `hedera-cryptography-wraps` directory today as it will only build the wraps dynamic library, and would require a manual run of Cargo in order to build the ceremony binary (although I may be wrong.) And the reason why we need this to work is because we're going to use artifacts produced by our release builds in GitHub that will need to be signed and deployed to CNs. So we shouldn't need a manual step to build this additional binary for that.

cryptography/hedera-cryptography-wraps/src/main/rust/wraps/src/preprocessing.rs
```diff
@@ -789,9 +789,10 @@ mod tests {
         }
     }
 
-    const MIMC_ROUNDS: usize = 3222;
-    const USE_MPC_CEREMONY: bool = true;
-    const PREPROCESS_CIRCUIT: bool = true;
+    const MIMC_ROUNDS: usize = 32222;
```
anthony-swirldslabs: Minor: this is an unusual value for a constant. It would be nice to add a comment above it in the code and explain in a few words what `32222` actually means.

cryptography/hedera-cryptography-wraps/src/main/rust/wraps/Cargo.toml
```diff
@@ -6,7 +6,7 @@ edition.workspace = true
 [lib]
 name = "wraps"
 path = "src/lib.rs"
-crate-type = ["cdylib"]
+crate-type = ["cdylib", "rlib"]
```
anthony-swirldslabs, line 9: Minor: to add to my above comment, I'm wondering how does Cargo decide which type to apply to which artifact?

cryptography/hedera-cryptography-wraps/src/main/rust/wraps/src/preprocessing.rs
rsinha: Ah, this is just used for the test circuit, and I vary it to make the circuit larger / smaller. Let me add a comment.

cryptography/hedera-cryptography-wraps/src/main/rust/wraps/Cargo.toml
rsinha: I believe that the cli, which is a binary artifact, will use the `rlib`, and the wraps artifact, which is a library, will use the `cdylib`.

cryptography/hedera-cryptography-wraps/src/main/rust/ceremony/src/main.rs
rsinha: Ok I can make the change.

cryptography/hedera-cryptography-wraps/src/main/rust/ceremony/Cargo.toml
rsinha: Makes sense, but I might need your guidance on how to include it in your gradle build.

anthony-swirldslabs: Per my understanding, once we have the -ceremony/src/main/rust/ populated with the code that implements our ceremony tool in full, we just need to add the above linked line to the already existing `hedera-cryptography-ceremony/build.gradle.kts`, as well as list the `org.hiero.gradle.feature.rust` plugin at its very top. Technically, that should be all that's needed.

The trickiest part would be to see how to properly extract the required Rust code, so that we minimize (or ideally, completely avoid) any code duplication between -wraps and -ceremony.

The less tricky but still untested part, is to see how our Rust Gradle support works when we're building an `org.hiero.gradle.module.application` as opposed to `org.hiero.gradle.module.library` - that's the only major difference between the -ceremony and -wraps modules.

cryptography/hedera-cryptography-wraps/src/main/rust/wraps/Cargo.toml
anthony-swirldslabs: Yeah, I get this. My question is how does Cargo figure out which to apply to which build? The code here doesn't really provide any hints on that. If I read it as is, I'd assume it'd build both the `cdylib` and `rlib` variants for both the projects (but again, I may be wrong. My level of expertise in Cargo is very close to zero.)

Anyway, as long as the Cargo.toml in this PR is able to still build the main wraps dynamic library and all Java unit tests run and pass accordingly, then it's all great. We'd probably be separating the tool from the library in the future per another comment here, so we can always revert this change here.

rsinha: Cargo has multiple targets: `ceremony` and `wraps`. The former forces the compilation under `rlib`  and the latter forces compilation under `cdylib`.

anthony-swirldslabs (approved):
LGTM. Thank you!

PR #525, opened 2026-02-11
feat: add jpackage support

### LINKED ISSUE #524: Support `jpackage` in Ceremony build

anthony-swirldslabs:
**Description**:
Adding support for `./gradlew jpackage` which can produce installable packages for target operating systems.

I was able to produce a .dmg package on my Mac by running: `./gradlew jpackageDarwin-aarch64` and it can be installed, and then I can simply run it as:

```shell
$ /Applications/hedera-cryptography-ceremony.app/Contents/MacOS/hedera-cryptography-ceremony
Usage: Orchestrator thisNodeId nodeId1,nodeId2,... dataCruncherExecutableFile s3Region s3Endpoint s3BucketName keyStorePath keyStorePassword
```

(the usage output is printed by our Orchestrator class.)

**Related issue(s)**:

Fixes #524 

**Notes for reviewer**:
This is a manually run target, so its presence shouldn't affect anything already setup in the build. All tests should pass.

**Checklist**

### DIFF
--- a/cryptography/hedera-cryptography-ceremony/build.gradle.kts
+++ b/cryptography/hedera-cryptography-ceremony/build.gradle.kts
@@ -36,6 +36,7 @@ javaModulePackaging {
         architecture = MachineArchitecture.X86_64
     }
 
-    // Bouncy Castle JARs fail as: jlink failed with: Error: signed modular JAR is currently not supported:
-    jlinkOptions.addAll("--ignore-signing-information");
+    // Bouncy Castle JARs fail as: jlink failed with: Error: signed modular JAR is currently not
+    // supported:
+    jlinkOptions.addAll("--ignore-signing-information")
 }
--- a/cryptography/hedera-cryptography-ceremony/build.gradle.kts
+++ b/cryptography/hedera-cryptography-ceremony/build.gradle.kts
@@ -2,6 +2,7 @@
 plugins {
     id("org.hiero.gradle.module.application")
     id("org.hiero.gradle.feature.shadow")
+    id("org.gradlex.java-module-packaging") version "1.2"
 }
 
 testModuleInfo {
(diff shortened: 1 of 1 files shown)


rsinha (approved):

cryptography/hedera-cryptography-ceremony/build.gradle.kts
```diff
@@ -12,3 +13,30 @@ testModuleInfo {
...
+        operatingSystem = OperatingSystemFamily.LINUX
+        architecture = MachineArchitecture.ARM64
+    }
+    target("darwin-aarch64") {
+        operatingSystem = OperatingSystemFamily.MACOS
+        architecture = MachineArchitecture.ARM64
+    }
+    target("darwin-x86_64") {
+        operatingSystem = OperatingSystemFamily.MACOS
+        architecture = MachineArchitecture.X86_64
```
jjohannes, line 32: Just a note:

By default, the plugin builds all package types Java's jpackage currrently supports. In the case of MACOS, for example, `pkg` and `dmg`. If you want to reduce that, you could set `packageTypes = listOf("dmg")` ([full example](https://github.com/jjohannes/gradle-project-setup-howto/blob/java_module_system/gradle/plugins/src/main/kotlin/org.example.gradle.feature.jpackage.gradle.kts#L4-L28))

I am afraid this is poorly documented. [These are the defaults](https://github.com/gradlex-org/java-module-packaging/blob/f566769e821e48f7d974608d2ee00033d749e4e5/src/main/java/org/gradlex/javamodule/packaging/JavaModulePackagingExtension.java#L186-L191).

jjohannes (approved):

cryptography/hedera-cryptography-ceremony/build.gradle.kts
anthony-swirldslabs: Thanks for the note! I think we're good with producing all the default package types for now. We can decide which we really need at a later time. Overall though, I found `java-module-packaging` very well documented.

PR #526, opened 2026-02-15
chore: ceremony cleanup

rsinha:
**Description**:
Based on testing, we think some optimizations are unnecessary.

### DIFF
--- a/cryptography/hedera-cryptography-wraps/src/main/rust/ceremony/src/main.rs
+++ b/cryptography/hedera-cryptography-wraps/src/main/rust/ceremony/src/main.rs
@@ -3,7 +3,7 @@ use std::path::PathBuf;
 
 use wraps::preprocessing::WRAPSPreprocessing;
 
-const NUM_COORDINATOR_THREADS: usize = 32;
+const NUM_COORDINATOR_THREADS: usize = 48;
 const NUM_PARTICIPANT_THREADS: usize = 4;
 
 fn setup_rayon_thread_pool(num_threads: usize) {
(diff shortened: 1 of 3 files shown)


cryptography/hedera-cryptography-wraps/src/main/rust/ceremony/src/main.rs
```diff
@@ -103,6 +113,7 @@ fn run(args: Args) -> Result<(), String> {
             Ok(())
         }
         2 => {
+            setup_rayon_thread_pool(NUM_PARTICIPANT_THREADS);
```
anthony-swirldslabs, line 116: Minor: if we ever add a global call to the setup() method, then these overridden calls will have no effect (as you told me, rayon can only be initialized once.)

Since we use the same num of threads in all cases, it may be safer to make this call just once regardless of the phase. This would avoid code duplication and ensure that we actually initialize rayon just once.

anthony-swirldslabs (dismissed):

cryptography/hedera-cryptography-wraps/src/main/rust/ceremony/src/main.rs
rsinha: @anthony-swirldslabs good point. Please see that the threads are different for coordinator vs. consensus nodes, so we have 2 constants.

anthony-swirldslabs: Oh, I see. The constant names look too very similar, so I didn't even notice that. If the values are indeed different, then the current approach looks good. Thanks for the clarification!

anthony-swirldslabs (dismissed):

anthony-swirldslabs (approved):

PR #538, opened 2026-02-20
fix: ceremony logic changes

rsinha:
Confirmed that with these changes, the TSS ceremony works end-to-end, where the resulting wraps prover and verification key works with the rest of library.


anthony-swirldslabs (approved):
LGTM. Thanks!

PR #555, opened 2026-03-03
perf: optimized deserialization of prover key

rsinha:
**Description**:
This PR uses an optimized loading of WRAPS prover key in memory. It turns off checks (whether the group elements lie on the elliptic curve), which is safe because we have a trusted source for the prover key.

This brings down the WRAPS proof generation down to under 10 minutes.


jsync-swirlds (approved):
LGTM, and thanks.

anthony-swirldslabs (approved):

PR #561, opened 2026-03-09
fix: overhaul ceremony timeouts

### LINKED ISSUE #560: Overhaul timeouts
Make timeouts more dynamic and "less" waiting.

anthony-swirldslabs:
**Description**:
Overhauling the timeout used to wait for a phase input to make it more dynamic and "less" waiting. Major change:
1. Wait for the desired input in 2 hours iterations.
2. On each iteration, consider reducing the total timeout if lower nodes have made some work.

**Related issue(s)**:

Fixes #560 

**Notes for reviewer**:
All tests should pass.

**Checklist**


cryptography/hedera-cryptography-ceremony/src/main/java/com/hedera/cryptography/ceremony/Phase.java
```diff
@@ -61,38 +64,79 @@ private String readyFileNameAtIndex(int index) {
...
-            } finally {
-                // So on the first iteration we waited for all the nodes. That's a very-very long wait.
-                // Subsequent iterations will use a single-node timeout because every previous node
-                // that could produce any data should've already produced it.
-                timeoutMillis = WAIT_FOR_REGULAR_FILE_TIMEOUT_MILLIS;
+            }
+            try {
+                Thread.sleep(WAIT_BETWEEN_CLAIMED_FILES_CHECKS_MILLIS);
+            } catch (InterruptedException e) {
+                // Swallow spurious exceptions.
```
jsync-swirlds, line 115: This could be `parkNanos(WAIT_BETWEEN_CLAIMED_FILES_CHECKS_MILLIS * 1000000L)` and avoid the exception check.

cryptography/hedera-cryptography-ceremony/src/main/java/com/hedera/cryptography/ceremony/Phase.java
```diff
@@ -61,38 +64,79 @@ private String readyFileNameAtIndex(int index) {
...
+                waitUntilMillis = Math.min(
+                        waitUntilMillis,
+                        System.currentTimeMillis()
+                                + WAIT_FOR_REGULAR_FILE_TIMEOUT_MILLIS
+                                        * Math.max(1, thisIndex - findMostRecentClaimedIndex(thisIndex)));
+
+                if (s3DirectoryAccessor.waitForFile(readyFileName, WAIT_ITERATION_MILLIS)) {
+                    System.err.println("Found file: " + readyFileName);
+                    return thisIndex - 1;
+                } else if ("666".equals(phase)) {
```
jsync-swirlds: Perhaps choose `3600` as a less provocative value?

cryptography/hedera-cryptography-ceremony/src/main/java/com/hedera/cryptography/ceremony/Phase.java
```diff
@@ -61,38 +64,79 @@ private String readyFileNameAtIndex(int index) {
...
+        System.err.println("Checking lower .claimed files to decide how long to wait...");
+        for (int i = thisIndex - 1; i >= 0; i--) {
+            if (s3DirectoryAccessor.doesExist(claimedFileNameAtIndex(i))) {
+                return i;
+            }
+            try {
+                Thread.sleep(WAIT_BETWEEN_CLAIMED_FILES_CHECKS_MILLIS);
+            } catch (InterruptedException e) {
+                // Swallow spurious exceptions.
+            }
```
jsync-swirlds, line 135: Same here, `parkNanos` doesn't check for interrupt or throw `InterruptedException`.

jsync-swirlds (dismissed):
A few minor nits; nothing blocking.

cryptography/hedera-cryptography-ceremony/src/main/java/com/hedera/cryptography/ceremony/Phase.java
anthony-swirldslabs: `LockSupport.parkNanos` has an extra logic related to "permits". Its javadoc suggests that it would return immediately if a permit is available, and it says a single permit is associated with each thread that uses this method. So if I read it correctly, it seems that the very first call to this method may not actually "park". The second and further should park. I'm unsure I want to deal with this extra complexity.

However, `Thread.sleep()` reads and feels natural.

cryptography/hedera-cryptography-ceremony/src/main/java/com/hedera/cryptography/ceremony/Phase.java
anthony-swirldslabs: Sure.

cryptography/hedera-cryptography-ceremony/src/main/java/com/hedera/cryptography/ceremony/Phase.java
anthony-swirldslabs: Same as above.

cryptography/hedera-cryptography-ceremony/src/main/java/com/hedera/cryptography/ceremony/Phase.java
jsync-swirlds: It's a bit simpler than it appears.
`Thread.sleep` calls `LockSupport.parkNanos` behind the scenes and checks for an interrupt after it returns.
Calling `parkNanos` is a pretty close equivalent to calling `sleep` and ignoring the exception; just without the unwanted exception catch.

It's a nitpick and you can leave as-is if you prefer to catch and ignore the interrupted exception.

anthony-swirldslabs: Perhaps. In the current JDK implementation that is.

Myself, I'm very much aligned with https://stackoverflow.com/a/10397946 - code readability matters.

rsinha (approved):

PR #564, opened 2026-03-11
fix: improve constraints

### LINKED ISSUE #563: Improve constraints
Misc improvements per suggestions.

anthony-swirldslabs:
**Description**:
Implementing a few improvements in constraint validations per suggestions from @dr20240304 , and removing an unused Gradle plugin.

**Related issue(s)**:

Fixes #563 

**Notes for reviewer**:
All tests should pass.

**Checklist**


common/hedera-common-nativesupport/src/main/java/com/hedera/common/nativesupport/api/StreamAPI.java
```diff
@@ -24,6 +24,11 @@ public abstract class StreamAPI implements ExternalAPI {
     /** @return an InputStream for the API to provide results to the client. */
     protected abstract InputStream getInputStream();
 
+    /** @return the max size of an array that receiveArray() can receive. */
+    protected int getMaxArraySize() {
```
rsinha, line 28: Does this mean we are allowing arrays of size Integer.MAX_VALUE?

anthony-swirldslabs: Yes, by default, which is 2GB. If an application wants a stricter size, they can override this method.

A side note: we don't use this class in the current version of tss-lib anywhere. So this particular patch is more of a protection for the future, if we ever use it that is.

rsinha (approved):

PR #565, opened 2026-03-13
fix: improved wraps error handling

rsinha:
**Description**:
This PR improves the error handling in WRAPS library. It removes the use of asserts and all unwraps that could be triggered during runtime.
Addresses https://github.com/hashgraph/hedera-cryptography/issues/467


cryptography/hedera-cryptography-wraps/src/main/rust/wraps/src/lib.rs
```diff
@@ -998,6 +1060,7 @@ impl WRAPS {
         hints_vk: impl AsRef<[u8]>,                          // TSS verification key for the next AddressBook
         multi_signature: &SchnorrMultiSignature,         // threshold Schnorr signature attesting the next AddressBook
     ) -> Result<(UncompressedProofSerialized, CompressedProofSerialized), WRAPSError> {
+        let invalid_input = |msg: &str| WRAPSError::InvalidInput(msg.to_string());
```
anthony-swirldslabs, line 1069: Minor: this could probably be a static symbol (maybe even a simple `fn`) to avoid duplicating it.

anthony-swirldslabs (dismissed):

cryptography/hedera-cryptography-wraps/src/main/rust/wraps/src/lib.rs
rsinha: Ah it gets inlined.

anthony-swirldslabs (approved):
LGTM

Issue #567, opened 2026-03-16
Secret Key Memory Zeroization Missing


rsinha:
The SecretKey type is defined as a simple type alias with no Drop implementation:
pub type SecretKey = F;  // Line 47 - just an alias to field element

PR #568, opened 2026-03-16
fix: zeroize secret key in memory

rsinha:
**Description**:
This PR adds a drop and zeroize trait to `SecretKey` in hinTS, so we clear secrets from app's memory after processing.

Addresses #567

### DIFF
--- a/cryptography/hedera-cryptography-hints/Cargo.toml
+++ b/cryptography/hedera-cryptography-hints/Cargo.toml
@@ -31,3 +31,4 @@ ark-serialize = { version = "^0.4.0", default-features = true }
 rand = { version = "^0.8.0" }
 rand_chacha = { version = "^0.3.0" }
 jni = "0.21.1"
+zeroize = "1.8"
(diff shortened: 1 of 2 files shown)


cryptography/hedera-cryptography-hints/Cargo.toml
```diff
@@ -31,3 +31,4 @@ ark-serialize = { version = "^0.4.0", default-features = true }
 rand = { version = "^0.8.0" }
 rand_chacha = { version = "^0.3.0" }
 jni = "0.21.1"
+zeroize = "1.8"
```
anthony-swirldslabs, line 34: Minor: do we have to take this new dependency? Cannot we simply run a simply for() loop to zero out the bytes?

anthony-swirldslabs (approved):

cryptography/hedera-cryptography-hints/Cargo.toml
rsinha: I believe it is idiomatic rust to use the trait in that crate. It also makes sure the compiler does not optimize it away.

PR #577, opened 2026-03-23
fix: safety checks in ceremony

rsinha:
**Description**:
This PR implements verification of each contribution in the ceremony.

### DIFF
--- a/cryptography/hedera-cryptography-wraps/src/main/rust/wraps/src/preprocessing.rs
+++ b/cryptography/hedera-cryptography-wraps/src/main/rust/wraps/src/preprocessing.rs
@@ -172,13 +172,15 @@ fn prove_knowledge_phase1(prev_srs: &PathBuf, next_srs: &PathBuf, tau: &Fr, alph
 }
 
 fn verify_knowledge_phase1(prev_srs_path: &PathBuf) {
-    let transcript = load_from_file::<PhasePokTranscript>(
+    let transcript = match load_from_file::<PhasePokTranscript>(
         &prev_srs_path.join("pok_transcript.bin")
-    ).unwrap();
-    if transcript.is_empty() {
-        println!("No proof of knowledge found for phase 1. Skipping verification.");
-        return;
-    }
+    ) {
+        Ok(t) if !t.is_empty() => t,
+        _ => {
+            println!("No proof of knowledge found for phase 1. Skipping verification.");
+            return;
+        }
+    };
     let tx = transcript.last().unwrap();
 
     let transcript_prev = transcript
--- a/cryptography/hedera-cryptography-wraps/src/main/rust/wraps/src/preprocessing.rs
+++ b/cryptography/hedera-cryptography-wraps/src/main/rust/wraps/src/preprocessing.rs
@@ -2,11 +2,11 @@
 
 use std::path::PathBuf;
 use std::io::Write;
-use ark_bn254::{G1Affine, G2Affine, g1};
-use ark_ec::{CurveGroup, AffineRepr};
+use ark_bn254::{G1Affine, G2Affine, g1, Bn254 as Curve};
+use ark_ec::{CurveGroup, AffineRepr, pairing::Pairing};
 use ark_ff::Field;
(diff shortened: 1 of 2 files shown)


cryptography/hedera-cryptography-wraps/src/main/rust/wraps/src/preprocessing.rs
```diff
@@ -89,6 +89,34 @@ pub struct CircuitConfig {
     pub num_instance_variables: usize,
 }
 
+// inspired from https://alinush.github.io/groth16
+#[derive(Clone, CanonicalDeserialize, CanonicalSerialize)]
+pub struct Phase1ProofOfKnowledge {
```
anthony-swirldslabs, line 94: Minor: are the "phase1" (and "phase2" below) actually refer to our Ceremony's phases 2 and 4? If so, it would be nice to use a consistent numbering to avoid any confusion in the future.

cryptography/hedera-cryptography-wraps/src/main/rust/wraps/src/preprocessing.rs
```diff
@@ -106,6 +134,192 @@ fn store_to_file<T: CanonicalSerialize>(path: &PathBuf, data: &T) -> Result<(),
...
+    ) {
+        Ok(t) if !t.is_empty() => t,
+        _ => Vec::new()
+    };
+    transcript_prev.push(proof_of_knowledge);
+    store_to_file::<Phase1PokTranscript>(&next_srs.join("pok_transcript.bin"), &transcript_prev).unwrap();
+}
+
+//prove_knowledge_phase1(prev_srs, next_srs, &tau, &alpha, &beta);
+fn prove_knowledge_phase2(prev_srs: &PathBuf, next_srs: &PathBuf, delta: &Fr, gamma: &Fr) {
```
anthony-swirldslabs: Minor: comment typo at line 213: `s/1/2/`. Not sure if this comment if helpful at all though.

anthony-swirldslabs (dismissed):
LGTM. Just a couple minor comments.

cryptography/hedera-cryptography-wraps/src/main/rust/wraps/src/preprocessing.rs
rsinha: True. Although the academic literature uses phase1 and phase2, so it might be clearer to keep that name for external readers. wdyt?

cryptography/hedera-cryptography-wraps/src/main/rust/wraps/src/preprocessing.rs
rsinha: I will remove that comment -- that was unnecessary.

rsinha: Fixed

cryptography/hedera-cryptography-wraps/src/main/rust/wraps/src/preprocessing.rs
anthony-swirldslabs: This is why (w/o realizing the academic reference though) I didn't want to rename phase1/2 to phase2/4 in the Ceremony Orchestrator. I'd rather you invent some special names for the pre-, post-, and interim- phases in the ceremony Rust executable instead.

But what's done is done. I'm okay with the current code, but I still find it confusing that we use different numbering of phases in different places. Especially in this very code that uses both at once - one when parsing command line arguments, and a different one here in all these function names in this entire file.

However, since it's unlikely that we'll need to run the ceremony more than once, I guess it's okay (not great, but okay) to leave things as they are for now.

anthony-swirldslabs (approved):

cryptography/hedera-cryptography-wraps/src/main/rust/wraps/src/preprocessing.rs
rsinha: That is true -- I agree. But we would need to support non-integer cli arguments for that, right?

PR #578, opened 2026-03-30
fix: ceremony verification logic

rsinha:
**Description**:
This PR implements additional security checks in the ceremony that can only be invoked by the coordinator.

### DIFF
--- a/cryptography/hedera-cryptography-wraps/src/main/rust/wraps/src/preprocessing.rs
+++ b/cryptography/hedera-cryptography-wraps/src/main/rust/wraps/src/preprocessing.rs
@@ -319,7 +319,7 @@ fn verify_knowledge_phase2(prev_srs_path: &PathBuf) {
 }
 
 // verication of consecutive steps in the ceremony -- this can only be done by the coordinator
-fn coordinator_verification(prev_srs_path: &PathBuf, next_srs_path: &PathBuf) {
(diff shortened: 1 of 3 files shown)


cryptography/hedera-cryptography-wraps/src/main/rust/wraps/src/preprocessing.rs
```diff
@@ -191,7 +191,7 @@ fn prove_knowledge_phase1(prev_srs: &PathBuf, next_srs: &PathBuf, tau: &Fr, alph
     let proof_of_knowledge = Phase1ProofOfKnowledge {
         pok_tau,
         node_contribution_tau: g_pow_tau,
-        next_tau: next_powers_of_tau_g1[0],
+        next_tau: next_powers_of_tau_g1[1],
```
anthony-swirldslabs, line 194: Minor: this seems like some sort of a bug fix? If so, then we'll need to make sure that the artifacts produced by the ceremony are still usable by WRAPS after this fix.

anthony-swirldslabs (approved):

cryptography/hedera-cryptography-wraps/src/main/rust/wraps/src/preprocessing.rs
rsinha: It is just fixing the proof of knowledge, which will only be used by the coordinator. It became relevant only with the new checks.

PR #580, opened 2026-04-13
chore: adding lock files for pinning cargo deps

rsinha:
**Description**:
This PR introduces lock files for pinning rust dependencies, which is necessary for open-sourcing.


cryptography/hedera-cryptography-hints/Cargo.lock
```diff
@@ -0,0 +1,678 @@
...
+checksum = "1a87aa2bb7d2af34197c04845522473242e1aa17c12f4935d5856491a7fb8c97"
+dependencies = [
+ "cesu8",
+ "cfg-if",
+ "combine",
+ "jni-sys 0.3.1",
+ "log",
+ "thiserror",
+ "walkdir",
+ "windows-sys 0.45.0",
```
anthony-swirldslabs, line 293: Minor: it's a bit odd to have this Windows-platform-specific dependency and not have similar dependencies for Linux and MacOS which we also support. But maybe this is just a quirk of the `jni` package.

anthony-swirldslabs (dismissed):

cryptography/hedera-cryptography-hints/Cargo.lock
rsinha: Yah that is strange.

rsinha: Not sure what we can do here. If we use jni, we have to live with it I think.

anthony-swirldslabs (approved):

PR #584, opened 2026-04-27
fix: pruning dependencies with multiple versions

rsinha:
**Description**:
This PR fixes the issue where `ark-std` was being used twice, with different commit hashes.

### DIFF
--- a/cryptography/hedera-cryptography-wraps/Cargo.toml
+++ b/cryptography/hedera-cryptography-wraps/Cargo.toml
@@ -20,7 +20,7 @@ ark-relations = { git = "https://github.com/arkworks-rs/snark" }
 ark-snark = { git = "https://github.com/arkworks-rs/snark" }
 ark-crypto-primitives = { git = "https://github.com/winderica/crypto-primitives", rev = "af003fc" }
 ark-r1cs-std = { git = "https://github.com/winderica/r1cs-std", rev = "ae8283a" }         # "sw-fix-updated" branch
-ark-std = { git = "https://github.com/arkworks-rs/std" }
+ark-std = { git = "https://github.com/arkworks-rs/std", rev = "73dcad0", default-features = false }
 ark-poly-commit = { git = "https://github.com/arkworks-rs/poly-commit" }
 
 
--- a/cryptography/hedera-cryptography-wraps/Cargo.lock
+++ b/cryptography/hedera-cryptography-wraps/Cargo.lock
@@ -78,7 +78,7 @@ dependencies = [
  "ark-ec",
  "ark-ff",
  "ark-r1cs-std",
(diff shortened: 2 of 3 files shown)


hendrikebbers (dismissed):

cryptography/hedera-cryptography-wraps/Cargo.toml
```diff
@@ -20,7 +20,7 @@ ark-relations = { git = "https://github.com/arkworks-rs/snark" }
 ark-snark = { git = "https://github.com/arkworks-rs/snark" }
 ark-crypto-primitives = { git = "https://github.com/winderica/crypto-primitives", rev = "af003fc" }
 ark-r1cs-std = { git = "https://github.com/winderica/r1cs-std", rev = "ae8283a" }         # "sw-fix-updated" branch
-ark-std = { git = "https://github.com/arkworks-rs/std" }
+ark-std = { git = "https://github.com/arkworks-rs/std", rev = "73dcad0", default-features = false }
```
hendrikebbers, line 23: Why do we have a different behavior (git url + rev) for `ark-std` but the other ark libs we do it like `ark-snark = { version = "0.5.0", default-features = false }`
Is `ark-std` not published to cargo?

rsinha: I wanted to fix the commit hash. Does that also work for cargo?

anthony-swirldslabs: In `cryptography/hedera-cryptography-wraps/src/main/rust/groth16/Cargo.toml` at line 26 above I see

```
ark-std = { version = "0.5.0", default-features = false }
```

and here we pull it from github instead.

Can we use a single consistent source for `ark-std` everywhere in this library? I don't mind which particular variant we use - cargo or github - I just want it to be consistent for this (and any other) dependency.

rsinha: There is a patch in root Cargo.toml, and the resolver uses that to determine the `ark-std` version. The one in groth16 is mainly there from the original project that we "forked".

rsinha: As Cargo.lock shows, there is only a single version of ark-std, the one specified by patch.

anthony-swirldslabs: Would anything break if we update the groth16 Cargo.toml above to also use

```
ark-std = { git = "https://github.com/arkworks-rs/std", rev = "73dcad0", default-features = false }
```

?

rsinha: No we could do that

anthony-swirldslabs (approved):

PR #585, opened 2026-04-27
feat: sentinel key generation

rsinha:
**Description**:
This PR introduces a new method in the TSS library, used to generate sentinel keys. The sentinel key is such that we get a public key such that no one knows the corresponding private key. This is achieved using the Eligator hash-to-curve method.


cryptography/hedera-cryptography-wraps/src/main/rust/wraps/src/lib.rs
```diff
@@ -578,11 +584,40 @@ fn generate_proof_of_knowledge(
...
+
+// Domain separator passed to `MapToCurveBasedHasher::new` when hashing
+// `SENTINEL_KEY_INPUT` to the JubJub curve.
+const SENTINEL_KEY_DST: &[u8] = b"HieroTSSWrapsSentinelKey";
+
+/// Hashes the fixed `SENTINEL_KEY_INPUT` bytes to a JubJub curve point using
+/// arkworks' `MapToCurveBasedHasher` with `Elligator2Map` (true hash-to-curve).
+/// The resulting point is in the prime-order subgroup (cofactor cleared) and is
+/// returned in the upstream `ark_ed_on_bn254::EdwardsAffine` type.
+fn sentinel_pubkey() -> SchnorrPubKey {
```
anthony-swirldslabs, line 604: CC: @tinker-michaelj 

Should these really be constants hard-coded in the library? Shouldn't we instead accept these as arguments and let the caller (e.g. the Execution, or whoever else is going to call this method) define the constants?

Otherwise it's unclear why we bother with generating it anew every time - we could just hard-code the public key itself directly and return it from here w/o any computations.

cryptography/hedera-cryptography-wraps/src/main/rust/wraps/src/lib.rs
```diff
@@ -578,11 +584,40 @@ fn generate_proof_of_knowledge(
...
+
 // Verifies a Schnorr proof of knowledge of the discrete log of the public key.
+// Sentinel keys (deterministically derived from `SENTINEL_KEY_INPUT`) bypass this
+// check because they are placeholders without a real signer.
 fn verify_proof_of_knowledge(
     pok: &SchnorrPoK,
     pubkey: &SchnorrPubKey
 ) -> bool {
+    if *pubkey == sentinel_pubkey() {
+        return true;
```
anthony-swirldslabs: The `true` return value looks a bit concerning. Doesn't this logic belong to the `verify_addressbook()` instead? Is it fair/correct/safe that the latter returns `true` if every and all keys in the AB are sentinel?

cryptography/hedera-cryptography-wraps/src/main/rust/wraps/src/lib.rs
```diff
@@ -686,6 +721,23 @@ impl WRAPS {
...
     }
 
+    /// Returns the deterministic sentinel `SchnorrAttestedPubKey`.
+    ///
+    /// The public key is derived by hashing a fixed byte string `SENTINEL_KEY_INPUT`
+    /// to the JubJub curve. The proof of knowledge is
+    /// populated with zero values; `verify_proof_of_knowledge` short-circuits
+    /// to `true` whenever the supplied public key matches this sentinel, so the
+    /// dummy PoK is never validated against the dlog relation.
+    pub fn sentinel_keygen() -> SchnorrAttestedPubKey {
```
anthony-swirldslabs, line 732: Minor: since this is only used in tests, could this function be moved to the `#[test]` ?

cryptography/hedera-cryptography-wraps/src/main/rust/wraps/src/lib.rs
rsinha: This function is supposed to return a deterministic value, and we need to hash some string to a curve element. I dont think the caller should be responsible, if it is just as well a constant.

cryptography/hedera-cryptography-wraps/src/main/rust/wraps/src/lib.rs
rsinha: From TSS point of view, that would be a valid address book.

cryptography/hedera-cryptography-wraps/src/main/rust/wraps/src/lib.rs
rsinha: This is a new method to be called by execution layer whenever a node does not supply its Schnorr pubkey.

cryptography/hedera-cryptography-wraps/src/main/rust/wraps/src/lib.rs
rsinha: But I agree it belongs there. Making the chnge.

cryptography/hedera-cryptography-wraps/src/main/rust/wraps/src/lib.rs
rsinha: @anthony-swirldslabs I believe this would need a new method to be added to jni.

anthony-swirldslabs (approved):
LGTM. Thanks!

cryptography/hedera-cryptography-wraps/src/main/rust/wraps/src/lib.rs
anthony-swirldslabs: Sounds good.

cryptography/hedera-cryptography-wraps/src/main/rust/wraps/src/lib.rs
anthony-swirldslabs: Just pushed a commit with the API.

PR #607, opened 2026-05-19
docs: adding whitepaper

rsinha:
**Description**:
This PR adds the whitepaper describing HieroTSS, prepared by @pratyay85 .


README.md
```diff
@@ -8,6 +8,8 @@
 
 This section is a work in progress.
 
+The cryptographic protocol description is contained within [this file](./docs/whitepaper.pdf).
```
anthony-swirldslabs, line 11: It mentions a "192-byte proof" for the compressed proof, but actually it's 704 bytes: https://github.com/hashgraph/hedera-cryptography/blob/main/cryptography/hedera-cryptography-wraps/src/main/java/com/hedera/cryptography/tss/TSS.java#L15


How come?

Also, the paper mentions "Hedera". This looks a bit confusing because it seems to be used interchangeably with "Hiero". We might want to s/Hedera/Hiero/ in the .pdf, or if we must mention Hedera, then it should be mentioned in the context of "Hedera uses Hiero software as an implementation of the ledger system" or something like that.

rsinha: Good points. @pratyay85 do you want to make the edits?

pratyay85: Good point, the Groth16 proof is 192 bytes, but the TSS also contains the current hinTS which is around 512 Bytes. I will also changed the mention of Hedera in the whitepaper.

pratyay85 (dismissed):
I updated the overleaf, @rsinha please change the pdf accordingly. Thanks!

README.md
anthony-swirldslabs: @pratyay85 Thanks for fixing these issues. The document looks good!

anthony-swirldslabs (dismissed):

anthony-swirldslabs (approved):
I'm unsure what exactly changed in the .pdf because GitHub won't show a diff for binary files. But overall, the doc still looks good.

I'll let @rsinha check the specific changes in the doc.

pratyay85 (approved):
Looks good

rbarker-dev (changes requested):
We generally do not store pdfs in github

PR #612, opened 2026-06-11
build: activate Gradle project isolation for faster build startup

jjohannes:
**Description**:

See: https://github.com/hiero-ledger/hiero-gradle-conventions/issues/477

**Related issue(s)**:

Updates convention plugins for:

- https://github.com/hiero-ledger/hiero-gradle-conventions/issues/477

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -3,4 +3,4 @@ org.gradle.jvmargs=-Xmx2048m
 
 # Enable Gradle Caching
 org.gradle.caching=true
-org.gradle.configuration-cache=true
+org.gradle.unsafe.isolated-projects=true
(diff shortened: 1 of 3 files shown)


gradle.properties
```diff
@@ -3,4 +3,4 @@ org.gradle.jvmargs=-Xmx2048m
 
 # Enable Gradle Caching
 org.gradle.caching=true
-org.gradle.configuration-cache=true
+org.gradle.unsafe.isolated-projects=true
```
anthony-swirldslabs, line 6: I barely understand this Gradle feature, but the change looks okay in general.

I'm a little concerned about the word `unsafe` in this option name though. I'd rather much prefer not to use anything that's marked "unsafe", "experimental", etc. If we must use anything that's marked like this, it's best to add a comment directly in code right above this line and justify the usage (not just say "it's faster", but something like "it's safe in our case because blah-blah".)

Regardless, I'd approve this PR, but the build seems to be broken - it seems to try to build Mac binaries on a Linux host or something. Please check the PR checks.

PR #628, opened 2026-06-17
build: overhaul libsodium cross-platform aggregation logic

### LINKED ISSUE #623: Revisit libsodium cross-platform build logic
Per suggestions from @jjohannes , consider hard-coding supported targets as separate tasks in Gradle and then rely on the Gradle cache and the `jar` task dependency to aggregate all the artifacts together into the production jar.

anthony-swirldslabs:
**Description**:
Fixing the cross-platform aggregation of native libsodium libraries by means of setting up a cross-platform build on every host: both Linux and Mac runners in CI build libsodium for all platforms locally. This is done so because a Gradle Cache-based solution that we use for our hints and wraps Rust libraries didn't work for libsodium for an unknown reason.

Note: on Mac, libsodium doesn't require the Apple SDK which has licensing restrictions that force us to build hints and wraps on Mac natively. Since the Apple SDK isn't needed for libsodium, we can cross-compile libsodium for Mac on Linux.

On developer laptops, local builds continue to build libsodium for the local host platform only, because that's the only locally-runnable binary. If anyone needs to reproduce the full CI build for all platforms, then they can simply define the `HIERO_LIBSODIUM_TARGETS` env var similar to our CI setup and install cross-build tools, similar to how we configure our CI runners (see GitHub scripts in this PR.)

**Related issue(s)**:

Fixes #623 

**Notes for reviewer**:
All tests should pass.

**Checklist**

### DIFF
--- a/cryptography/libsodium/build.gradle.kts
+++ b/cryptography/libsodium/build.gradle.kts
@@ -16,8 +16,11 @@ tasks.test {
 }
 
 /// Where we check out the libsodium repo from GitHub into the local build/ directory:
-val libRepositoryDir = layout.buildDirectory.dir("third-party/libsodium")
-val libOutputDir = layout.buildDirectory.dir("third-party/libsodium/output")
+/// Must end with "libsodium" or whatever name the GitHub repo has:
+val libRepositoryDir = layout.buildDirectory.dir("libsodium/input/libsodium")
+/// Where build tasks write output to:
+/// Must be outside of input/ above so that Gradle is happy:
+val libOutputDir = layout.buildDirectory.dir("libsodium/output")
 
 tasks.register<GitClone>("cloneLibsodium") {
     localCloneDirectory = libRepositoryDir
--- a/cryptography/libsodium/build.gradle.kts
+++ b/cryptography/libsodium/build.gradle.kts
@@ -1,6 +1,7 @@
 // SPDX-License-Identifier: Apache-2.0
-import java.util.regex.Pattern
 import org.gradle.api.internal.file.FileOperations
+import org.gradle.internal.extensions.stdlib.capitalized
+import org.gradle.kotlin.dsl.register
 import org.hiero.gradle.tasks.GitClone
 
 plugins { id("org.hiero.gradle.module.library") }
@@ -8,14 +9,12 @@ plugins { id("org.hiero.gradle.module.library") }
 testModuleInfo { requires("org.junit.jupiter.api") }
 
 tasks.test {
-    // It's a bit odd why test doesn't depend on assemble by default...
-    dependsOn("assemble")
-
     jvmArgs(
         "--enable-native-access=com.hedera.common.nativesupport,com.hedera.cryptography.libsodium"
     )
 }
 
+/// Where we check out the libsodium repo from GitHub into the local build/ directory:
 val libDir = layout.buildDirectory.dir("libsodium")
 
 tasks.register<GitClone>("cloneLibsodium") {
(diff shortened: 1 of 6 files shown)


cryptography/libsodium/build.gradle.kts
```diff
@@ -25,110 +24,72 @@ tasks.register<GitClone>("cloneLibsodium") {
...
-            )
-            .get()
-    val targetOsArch = target.split(Pattern.compile("-"), 2)
+val mutexService =
+    gradle.sharedServices.registerIfAbsent("taskMutex", TaskMutex::class.java) {
+        maxParallelUsages.set(1)
+    }
+
+/// Builds a native library via ./configure && make and copies .so/.dylib/.dll to resources.
+abstract class BuildLibsodiumTask : DefaultTask() {
```
jjohannes, line 37:
```suggestion
@CacheableTask
abstract class BuildLibsodiumTask : DefaultTask() {
```

To have the results written to the remote cache.

cryptography/libsodium/build.gradle.kts
```diff
@@ -140,11 +101,114 @@ tasks.assemble {
...
+        }
+        .orElse(listOf(NativeTarget(hostOperatingSystem, hostArchitecture, "")))
+        .get()
+
+targets.forEach { target ->
+    val name = "buildLibsodium" + target.os.capitalized() + target.arch.capitalized()
+    println("Registering $name per $target")
+    val task =
+        tasks.register<BuildLibsodiumTask>(name) {
+            usesService(mutexService)
```
jjohannes: Good solution with the service!

Two things you _could_ do:

- You could reuse [`org.hiero.gradle.services.TaskLockService`](https://github.com/hiero-ledger/hiero-gradle-conventions/blob/main/src/main/kotlin/org/hiero/gradle/services/TaskLockService.kt) like [here](https://github.com/hiero-ledger/hiero-gradle-conventions/blob/main/src/main/kotlin/org.hiero.gradle.feature.test-timing-sensitive.gradle.kts#L15-L17). 
Then you can remove the `TaskMutex` class and change the service registration in this file.
```
gradle.sharedServices.registerIfAbsent("lock", TaskLockService::class) {
    maxParallelUsages = 1
}
``` 
 
- Instead of the line above, you could add this property to the BuildLibsodiumTask class:
```
    @get:ServiceReference("lock")
    abstract val lock : Property<TaskLockService>
```

cryptography/libsodium/build.gradle.kts
```diff
@@ -140,11 +101,114 @@ tasks.assemble {
...
+                    )
+                }
+                .toList()
+        }
+        .orElse(listOf(NativeTarget(hostOperatingSystem, hostArchitecture, "")))
+        .get()
+
+targets.forEach { target ->
+    val name = "buildLibsodium" + target.os.capitalized() + target.arch.capitalized()
+    println("Registering $name per $target")
```
jjohannes: As this is configuration time, meaning this "runs" everytime you start Gradle or import into IDE, it would be good to remove this println once everything works.

(Or change to `logger.info()`)

cryptography/libsodium/build.gradle.kts
```diff
@@ -140,11 +101,114 @@ tasks.assemble {
...
+        }
+
+    // Include all built native libraries into the .jar:
+    tasks.jar { from(task) }
+
+    // The compile tasks must also depend on the native builds. Technically, only the local host
+    // target library is required, e.g. to run the code and/or unit tests locally.
+    // But Gradle requires all of them because they all make it into the resources/. So:
+    tasks.compileJava { dependsOn(name) }
+    tasks.compileTestJava { dependsOn(name) }
```
jjohannes: If I read the comment correctly, the issue is that the lib is required at **test runtime**. The Java code should not need the library to compile.

The issue is likely that the tests use the folders from `build/` directly and not the packaged `Jar` files.

Instead of configuring the Jar task, you can tell that the output of your task is part of the **resources** of the main source set. [We do the same for rust.](https://github.com/hiero-ledger/hiero-gradle-conventions/blob/main/src/main/kotlin/org/hiero/gradle/extensions/CargoExtension.kt#L93)


```suggestion
sourceSets["main"].resources.srcDir(task)
```

This should lead to the file packaged into the Jar (i.e. does the `jar {from(...)}` thing for you) and also put the folder your task produces of the runtime path of the tests.

cryptography/libsodium/build.gradle.kts
```diff
@@ -140,11 +101,114 @@ tasks.assemble {
...
+}
+
+tasks.named<Jar>("jar") {
+    // Each native library goes into its own os/arch/ subdirectory, and our `ls -lR` proves it.
+    // However, Gradle still complains:
+    //    Entry libsodium.dylib is a duplicate but no duplicate handling strategy has been set.
+    // So we do this:
+    duplicatesStrategy = DuplicatesStrategy.INCLUDE
+}
```
jjohannes:
```suggestion
```

You should be able to remove this, after the other comments are addressed.

cryptography/libsodium/build.gradle.kts
```diff
@@ -140,11 +101,114 @@ tasks.assemble {
...
+    println("Registering $name per $target")
+    val task =
+        tasks.register<BuildLibsodiumTask>(name) {
+            usesService(mutexService)
+            dependsOn("cloneLibsodium")
+            libraryDir = libDir
+            configureHost = target.configureHost
+            outputDir =
+                layout.buildDirectory.dir(
+                    "resources/main/com/hedera/nativelib/libsodium/${target.os}/${target.arch}"
```
jjohannes: `resources/main` is the output of another task already: `processResources` that copies all resources into a single location.

In order for the caching and incremental building to work, each task needs an isolated output directory. Change this to a unique location. E.g.:

```suggestion
                    "libsodium/${target.os}-${target.arch}"
```

This will:
- be picked up by `processResources` after the change proposed in a comment below –`sourceSets["main"].resources.srcDir(task)` – which will then "merge" your generated resources with other resources into `build/resources/main`
- Solve the `duplicatesStrategy` issue. This is currently an issue because it looks to Gradle as two tasks produced the same files.

In order to get the lib into the tight sub-directory (pakckage) you need to put it there inside your task action. So in that code further up do something like:

```
val targetDir = dstDir.get().dir("com/hedera/nativelib/libsodium/").dir(targetOsArch[0]).dir(targetOsArch[1])
```

cryptography/libsodium/build.gradle.kts
```diff
@@ -140,11 +101,114 @@ tasks.assemble {
...
+        .get()
+
+targets.forEach { target ->
+    val name = "buildLibsodium" + target.os.capitalized() + target.arch.capitalized()
+    println("Registering $name per $target")
+    val task =
+        tasks.register<BuildLibsodiumTask>(name) {
+            usesService(mutexService)
+            dependsOn("cloneLibsodium")
+            libraryDir = libDir
```
jjohannes: This can be written in one statement, if you tell that you need the result of the other task rather than using a plain directory constant:


```suggestion
            libraryDir = tasks.named<GitClone>("cloneLibsodium").flatMap { it.localCloneDirectory }
```

jjohannes (changes requested):

cryptography/libsodium/build.gradle.kts
anthony-swirldslabs: Done.

cryptography/libsodium/build.gradle.kts
anthony-swirldslabs: Done.

cryptography/libsodium/build.gradle.kts
anthony-swirldslabs: Done.

cryptography/libsodium/build.gradle.kts
anthony-swirldslabs: Done.

cryptography/libsodium/build.gradle.kts
anthony-swirldslabs: Done.

cryptography/libsodium/build.gradle.kts
anthony-swirldslabs: Done.

cryptography/libsodium/build.gradle.kts
anthony-swirldslabs: Done.

jjohannes (approved):
This looks great. Thanks @anthony-swirldslabs.

andrewb1269 (approved):

PR #630, opened 2026-06-23
ci: update how targets are handled in workflows

andrewb1269:
Update how the targets are handled in workflows.

Workflow for build test [here](https://github.com/hiero-ledger/hiero-cryptography/actions/runs/28051702696). 🏃

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -49,6 +49,7 @@ jobs:
         run: |
           TARGETS=$(grep '// HIERO_LIBSODIUM_TARGETS=' cryptography/libsodium/build.gradle.kts \
             | sed 's|.*HIERO_LIBSODIUM_TARGETS="\(.*\)"|\1|')
+          echo "Targets to build: ${TARGETS}"
           echo "HIERO_LIBSODIUM_TARGETS=${TARGETS}" >> "${GITHUB_ENV}"
 
       - name: Install go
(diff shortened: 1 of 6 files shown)


.github/workflows/build.yml
```diff
@@ -47,6 +45,13 @@ jobs:
       - name: Checkout Code
         uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
 
+      - name: Read libsodium CI targets from build file
+        run: |
+          TARGETS=$(grep 'HIERO_LIBSODIUM_TARGETS=' .github/workflows/support/.build-targets \
+            | sed 's|.*HIERO_LIBSODIUM_TARGETS="\(.*\)"|\1|')
+          echo "Targets to build: ${TARGETS}"
+          echo "HIERO_LIBSODIUM_TARGETS=${TARGETS}" >> "${GITHUB_ENV}"
```
anthony-swirldslabs: We'll be adding more native libraries. They will likely have the same targets. And we may want to rename the env var or maybe add separate vars. It would be nice not to have to go and revisit all the places where you use the global definition. Can this be fixed please?

andrewb1269: I can rename it to something else, would `HIERO_TARGETS` be a better var name?

.github/workflows/build.yml
```diff
@@ -30,14 +30,37 @@ env:
...
+      - name: Checkout Code
+        uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
+
+      - name: Read libsodium CI targets from targets file
+        id: read
+        run: |
+          source .github/workflows/support/.build-targets
+          echo "Targets to build: ${HIERO_LIBSODIUM_TARGETS}"
+          echo "libsodium-targets=${HIERO_LIBSODIUM_TARGETS}" >> "${GITHUB_OUTPUT}"
+
```
rbarker-dev, line 40: This should be its own callable flow that provides an output. Then you aren't reusing code everywhere.

rbarker-dev (changes requested):

.github/workflows/build.yml
andrewb1269: Switched to `workflow call` with a new workflow.

andrewb1269 (commented):
LGTM

rbarker-dev (approved):

PR #631, opened 2026-06-24
perf: benchmark libsodium

### LINKED ISSUE #622: Benchmark FFM libsodium against lazysodium
Once cross-platform build issues are resolved, write and run a bench in CN repo to compare the two.

anthony-swirldslabs:
**Description**:
A JMH benchmark to compare performance between our new Java FFM-based libsodium package in hiero-cryptography and the LazySodium library that is currently used in Consensus Node.

Per the results below, it appears that our libsodium is a tad faster than LazySodium. Also, it appears that `MemorySegment.ofArray()` is very lightweight, so we don't necessarily have to pre-cache MemorySegment instances (which doesn't make sense in general, at least for the message/signature arrays because they're dynamic by definition.) Finally, it seems that the NoChecks versions of the methods are indeed slightly faster, but the difference is insignificant. So it makes sense to continue to use the regular versions of the methods that add a layer of protection against passing invalid data to the native code.

Note that the numbers are obtained on a darwin-arm64 system. We don't have a ready-to-use facility for running hiero-cryptography benchmarks on Linux systems.

```
Benchmark                                                 Mode  Cnt   Score   Error   Units
LazysodiumBench.cryptoSignDetached                       thrpt   15  68.363 ± 1.207  ops/ms
LazysodiumBench.cryptoSignVerifyDetached                 thrpt   15  25.769 ± 0.217  ops/ms
```

```
Benchmark                                                        Mode  Cnt   Score   Error   Units
LibsodiumBench.cryptoSignDetached                               thrpt   15  74.249 ± 0.290  ops/ms
LibsodiumBench.cryptoSignDetachedNoChecks                       thrpt   15  74.214 ± 0.305  ops/ms
LibsodiumBench.cryptoSignDetachedNoChecks_MemorySegments        thrpt   15  74.586 ± 0.309  ops/ms
LibsodiumBench.cryptoSignDetached_MemorySegments                thrpt   15  74.592 ± 0.241  ops/ms
LibsodiumBench.cryptoSignVerifyDetached                         thrpt   15  26.667 ± 0.234  ops/ms
LibsodiumBench.cryptoSignVerifyDetachedNoChecks                 thrpt   15  26.805 ± 0.287  ops/ms
LibsodiumBench.cryptoSignVerifyDetachedNoChecks_MemorySegments  thrpt   15  27.187 ± 0.153  ops/ms
LibsodiumBench.cryptoSignVerifyDetached_MemorySegments          thrpt   15  26.741 ± 0.070  ops/ms
```

**Related issue(s)**:

Fixes #622 

**Notes for reviewer**:
See above.

**Checklist**

### DIFF
--- a/cryptography/libsodium/src/jmh/java/com/hedera/cryptogaphy/libsodium/LazysodiumBench.java
+++ b/cryptography/libsodium/src/jmh/java/com/hedera/cryptogaphy/libsodium/LazysodiumBench.java
@@ -1,0 +1,93 @@
+// SPDX-License-Identifier: Apache-2.0
+package com.hedera.cryptogaphy.libsodium;
+
+import com.goterl.lazysodium.LazySodiumJava;
+import com.goterl.lazysodium.SodiumJava;
+import com.goterl.lazysodium.interfaces.Sign;
+import java.util.Random;
+import java.util.concurrent.TimeUnit;
+import org.openjdk.jmh.annotations.Benchmark;
+import org.openjdk.jmh.annotations.BenchmarkMode;
+import org.openjdk.jmh.annotations.Fork;
+import org.openjdk.jmh.annotations.Level;
+import org.openjdk.jmh.annotations.Measurement;
+import org.openjdk.jmh.annotations.Mode;
+import org.openjdk.jmh.annotations.OperationsPerInvocation;
+import org.openjdk.jmh.annotations.OutputTimeUnit;
+import org.openjdk.jmh.annotations.Scope;
+import org.openjdk.jmh.annotations.Setup;
+import org.openjdk.jmh.annotations.State;
+import org.openjdk.jmh.annotations.TearDown;
+import org.openjdk.jmh.annotations.Warmup;
+import org.openjdk.jmh.infra.Blackhole;
+import org.openjdk.jmh.runner.Runner;
+import org.openjdk.jmh.runner.options.Options;
+import org.openjdk.jmh.runner.options.OptionsBuilder;
+
+/// A separate LazySodium bench to prevent loading both our and their native libs into the JVM.
+@SuppressWarnings("unused")
+@State(Scope.Benchmark)
+@Fork(3)
+@Warmup(iterations = 3)
+@Measurement(iterations = 5)
+@OutputTimeUnit(TimeUnit.MILLISECONDS)
+@BenchmarkMode(Mode.Throughput)
+public class LazysodiumBench {
+    private static final int INVOCATIONS = 10000;
+
+    static final Sign.Native LAZYSODIUM = new LazySodiumJava(new SodiumJava());
+
+    /// Non-deterministic because JMH is multithreaded and non-deterministic.
+    /// The non-determinism shouldn't affect the benchmark.
+    private static final Random RANDOM = new Random();
+
+    @State(Scope.Thread)
+    public static class LazysodiumState {
+        byte[] msg;
+        byte[] pk;
+        byte[] sk;
+        byte[] sig;
+
+        @Setup(Level.Trial)
+        public void setup() throws Throwable {
+            msg = new byte[64];
+            RANDOM.nextBytes(msg);
+
+            pk = new byte[32];
+            sk = new byte[64];
+            LAZYSODIUM.cryptoSignKeypair(pk, sk);
+
+            sig = new byte[64];
+
+            LAZYSODIUM.cryptoSignDetached(sig, msg, msg.length, sk);
+        }
+
+        @TearDown(Level.Trial)
+        public void tearDown() {}
+    }
+
+    @Benchmark
+    @OperationsPerInvocation(INVOCATIONS)
+    public void cryptoSignDetached(final LazysodiumState state, final Blackhole blackhole) throws Throwable {
+        for (int i = 0; i < INVOCATIONS; i++) {
+            blackhole.consume(LAZYSODIUM.cryptoSignDetached(state.sig, state.msg, state.msg.length, state.sk));
+        }
+    }
+
+    @Benchmark
+    @OperationsPerInvocation(INVOCATIONS)
+    public void cryptoSignVerifyDetached(final LazysodiumState state, final Blackhole blackhole) throws Throwable {
+        for (int i = 0; i < INVOCATIONS; i++) {
+            blackhole.consume(LAZYSODIUM.cryptoSignVerifyDetached(state.sig, state.msg, state.msg.length, state.pk));
+        }
+    }
+
+    public static void main(String[] args) throws Exception {
+        Options opt = new OptionsBuilder()
+                .include(LazysodiumBench.class.getSimpleName())
+                .jvmArgs("--enable-native-access=ALL-UNNAMED")
+                .build();
+
+        new Runner(opt).run();
+    }
+}
--- a/cryptography/libsodium/build.gradle.kts
+++ b/cryptography/libsodium/build.gradle.kts
@@ -5,7 +5,10 @@ import org.gradle.kotlin.dsl.register
 import org.hiero.gradle.services.TaskLockService
 import org.hiero.gradle.tasks.GitClone
 
-plugins { id("org.hiero.gradle.module.library") }
+plugins {
+    id("org.hiero.gradle.module.library")
+    id("org.hiero.gradle.feature.benchmark")
+}
 
 testModuleInfo { requires("org.junit.jupiter.api") }
 
@@ -13,6 +16,11 @@ tasks.test {
     jvmArgs(
         "--enable-native-access=com.hedera.common.nativesupport,com.hedera.cryptography.libsodium"
     )
+}
+
+jmhModuleInfo {
+    requires("com.hedera.cryptography.libsodium")
+    requires("com.goterl.lazysodium")
 }
 
 /// Where we check out the libsodium repo from GitHub into the local build/ directory:
--- a/cryptography/libsodium/build.gradle.kts
+++ b/cryptography/libsodium/build.gradle.kts
@@ -21,6 +21,7 @@ tasks.test {
 jmhModuleInfo {
     requires("com.hedera.cryptography.libsodium")
     requires("com.goterl.lazysodium")
+    requires("com.hedera.pbj.runtime")
 }
 
 /// Where we check out the libsodium repo from GitHub into the local build/ directory:
--- a/cryptography/libsodium/src/jmh/java/com/hedera/cryptogaphy/libsodium/LibsodiumBench.java
+++ b/cryptography/libsodium/src/jmh/java/com/hedera/cryptogaphy/libsodium/LibsodiumBench.java
@@ -1,0 +1,176 @@
+// SPDX-License-Identifier: Apache-2.0
+package com.hedera.cryptogaphy.libsodium;
+
+import com.hedera.cryptography.libsodium.Libsodium;
+import java.lang.foreign.MemorySegment;
+import java.util.Random;
+import java.util.concurrent.TimeUnit;
+import org.openjdk.jmh.annotations.Benchmark;
+import org.openjdk.jmh.annotations.BenchmarkMode;
+import org.openjdk.jmh.annotations.Fork;
+import org.openjdk.jmh.annotations.Level;
+import org.openjdk.jmh.annotations.Measurement;
+import org.openjdk.jmh.annotations.Mode;
+import org.openjdk.jmh.annotations.OperationsPerInvocation;
+import org.openjdk.jmh.annotations.OutputTimeUnit;
+import org.openjdk.jmh.annotations.Scope;
+import org.openjdk.jmh.annotations.Setup;
+import org.openjdk.jmh.annotations.State;
+import org.openjdk.jmh.annotations.TearDown;
+import org.openjdk.jmh.annotations.Warmup;
+import org.openjdk.jmh.infra.Blackhole;
+import org.openjdk.jmh.runner.Runner;
+import org.openjdk.jmh.runner.options.Options;
+import org.openjdk.jmh.runner.options.OptionsBuilder;
+
+@SuppressWarnings("unused")
+@State(Scope.Benchmark)
+@Fork(3)
+@Warmup(iterations = 3, time = 2)
+@Measurement(iterations = 5)
+@OutputTimeUnit(TimeUnit.MILLISECONDS)
+@BenchmarkMode(Mode.Throughput)
+public class LibsodiumBench {
+    private static final int INVOCATIONS = 10000;
+
+    private static final Libsodium LIBSODIUM = Libsodium.getInstance();
+
+    /// Non-deterministic because JMH is multithreaded and non-deterministic.
+    /// The non-determinism shouldn't affect the benchmark.
+    private static final Random RANDOM = new Random();
+
+    @State(Scope.Thread)
+    public static class LibsodiumState {
+        byte[] msg;
+        byte[] pk;
+        byte[] sk;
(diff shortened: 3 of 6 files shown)


cryptography/libsodium/src/jmh/java/com/hedera/cryptogaphy/libsodium/LazysodiumBench.java
```diff
@@ -0,0 +1,93 @@
+// SPDX-License-Identifier: Apache-2.0
+package com.hedera.cryptogaphy.libsodium;
```
abies: It should be cryptography probably

anthony-swirldslabs: Good catch! Fixed.

jasperpotts (approved):
One minor issue, the warmup config differs between the two classes — undermines the head-to-head comparison.
  This is the one real flaw for an A/B benchmark whose entire purpose is comparing the two libraries:
  - LazysodiumBench: @Warmup(iterations = 3) → 3 × 10s default = 30s warmup
  - LibsodiumBench: @Warmup(iterations = 3, time = 2) → 3 × 6s warmup

So far the benchmark numbers look great that the new code is faster than old 😀 I have one more concern that would be great to validate before making a switch to new code. That is how does new vs old compare in terms of garbage generated and GC pressure. The benchmark numbers can hide the fact the new one might do twice as many allocations as the old one. So it would be nice to double check this. Hopefully all is good 😀 I asked claude how best to do that and it seems there is a GC profiler built into JMH that can help.

**What is best way to compare memory allication and GC work caused by the two?**                                                                                                                                                                                                                                                                                              

> ⏺ The right tool is JMH's built-in GC profiler (-prof gc), and the one metric that matters most for this comparison is gc.alloc.rate.norm — normalized bytes allocated per operation. Unlike rates or GC counts, it's deterministic and JVM-tuning-independent, so it gives you a clean apples-to-apples number.
> 
>   How to enable it
> 
>   In each benchmark's main(), just add the profiler:
> 
>   import org.openjdk.jmh.profile.GCProfiler;
> 
>   Options opt = new OptionsBuilder()
>           .include(LibsodiumBench.class.getSimpleName())
>           .jvmArgs("--enable-native-access=ALL-UNNAMED")
>           .addProfiler(GCProfiler.class)   // <-- adds gc.alloc.rate.norm, gc.count, gc.time
>           .build();
> 
>   Or, for the Gradle path (the champeau me.champeau.jmh plugin this PR pulls in):
> 
>   jmh { profilers.add("gc") }
> 
>   Or ad-hoc on the CLI: -prof gc.
> 
>   What you get and how to read it
> 
>   Benchmark                              Mode  Cnt    Score   Units
>   LibsodiumBench.cryptoSignDetached      thrpt   15   74.2    ops/ms
>   …:·gc.alloc.rate.norm                  thrpt   15   48.0    B/op    ← the number to compare
>   …:·gc.alloc.rate                       thrpt   15  120.3    MB/sec
>   …:·gc.count                            thrpt   15   42      counts
>   …:·gc.time                             thrpt   15   85      ms
> 
>   - gc.alloc.rate.norm (B/op) — primary comparison. Heap bytes allocated per single sign/verify. The cached _MemorySegments variants should report ~0 B/op; the non-cached variants will report exactly the cost of the MemorySegment.ofArray() wrappers; LazySodium will report whatever JNA churns on the heap. This directly answers "who allocates more."
>   - gc.alloc.rate scales with throughput, so it's less fair across libraries with different op rates — use the normalized one.
>   - gc.count / gc.time — actual GC work (collections triggered, pause time). These are what you ultimately care about, but they're GC-config-dependent.
> 
>   Gotchas specific to this comparison
> 
>   1. Pin identical heap + GC settings in both classes, or gc.count/gc.time aren't comparable. Add the same jvmArgsAppend to both: e.g. @Fork(value = 3, jvmArgsAppend = {"-Xms2g", "-Xmx2g", "-XX:+UseG1GC"}). gc.alloc.rate.norm is immune to this (it's pure allocation), which is another reason to lead with it.
>   2. The GC profiler only sees the Java heap. This is the big one for FFM-vs-JNA: LazySodium/JNA allocates off-heap native temporary buffers (and Pointer/Memory objects), and the native malloc/free cost won't show up in gc.alloc.rate.norm at all. So the GC profiler is the right answer for "GC work", but it's not the whole "memory cost" story. The libsodium FFM
>   side with critical(true) pins heap arrays (no copy), so it should be near-zero heap and near-zero native temp allocation — that asymmetry is real and worth calling out, but the profiler will under-count JNA's native side.
>   3. Your Blackhole.consume calls are already allocation-safe — they hit the primitive consume(int) / consume(boolean) overloads, so no autoboxing noise pollutes the allocation numbers. Keep it that way (don't consume boxed types).
>   4. Don't run the GC-profiler trial on the cached vs non-cached question and trust tiny throughput deltas — gc.alloc.rate.norm is the honest way to show the difference. The throughput numbers in the PR are within error bars (allocation cost is dwarfed by the ~14µs native call), but gc.alloc.rate.norm will show a crisp, non-noisy B/op gap. That's a stronger,
>   more honest way to make the "do we need to pre-cache MemorySegments?" argument than the throughput numbers alone.
> 
>   If you want allocation call sites, not just totals
> 
>   Add a sampling profiler on top:
>   - -prof jfr → produces a JFR recording; open in JDK Mission Control to see allocation-by-type and GC pause distribution. Best for the "GC work" pause analysis.
>   - -prof async:libPath=...;output=flamegraph;event=alloc (async-profiler, works on darwin-arm64) → allocation flame graph showing exactly which call sites allocate. Best for proving what JNA allocates vs FFM.
> 
>   Bottom line: add -prof gc to both benchmarks (with identical heap/GC jvmArgs), and compare gc.alloc.rate.norm. Lead your write-up with that number, note explicitly that it captures heap-only (so JNA's native allocations are understated), and use -prof jfr or async-profiler alloc mode if you need to attribute the bytes to specific call sites.

anthony-swirldslabs:
Thanks, @jasperpotts !

After making the `@Warmup` configuration equal between the benchmarks (see the latest commit), and enabling the GCProfiler, below are the updated results.

Overall, the numbers are the same as obtained originally and posted in the description of this PR.

The new `gc.alloc.rate.norm` metric is ~275 B/op with Lazysodium and ~216 B/op with our FFM Libsodium for the `cryptoSignDetached` API. So I presume our FFM is doing better with less allocations. Note that if MemorySegments are cached, then the rate drops to zero. While this sounds great, and we should probably do that where possible, this may not be always possible - e.g. for the changing data such as signatures and hashes (aka messages being signed.) But even if don't do that, we still do better in Java heap, and as your comment suggests, we have even less allocations in native code. So it's a win-win even if don't cache the MemorySegments. For `cryptoSignVerifyDetached`, this metric is zero for LazySodium but still those 216 B/op for FFM Libsodium.

```
Benchmark                                                     Mode  Cnt     Score     Error   Units
LazysodiumBench.cryptoSignDetached                           thrpt   15    68.272 ±   0.891  ops/ms
LazysodiumBench.cryptoSignDetached:gc.alloc.rate             thrpt   15    17.904 ±   6.490  MB/sec
LazysodiumBench.cryptoSignDetached:gc.alloc.rate.norm        thrpt   15   275.734 ± 104.137    B/op
LazysodiumBench.cryptoSignDetached:gc.count                  thrpt   15    32.000            counts
LazysodiumBench.cryptoSignDetached:gc.time                   thrpt   15  1396.000                ms
LazysodiumBench.cryptoSignVerifyDetached                     thrpt   15    25.594 ±   0.220  ops/ms
LazysodiumBench.cryptoSignVerifyDetached:gc.alloc.rate       thrpt   15     0.001 ±   0.001  MB/sec
LazysodiumBench.cryptoSignVerifyDetached:gc.alloc.rate.norm  thrpt   15     0.027 ±   0.001    B/op
LazysodiumBench.cryptoSignVerifyDetached:gc.count            thrpt   15       ≈ 0            counts
```


```
Benchmark                                                                           Mode  Cnt    Score    Error   Units
LibsodiumBench.cryptoSignDetached                                                  thrpt   15   74.338 ±  0.250  ops/ms
LibsodiumBench.cryptoSignDetached:gc.alloc.rate                                    thrpt   15   15.313 ±  0.052  MB/sec
LibsodiumBench.cryptoSignDetached:gc.alloc.rate.norm                               thrpt   15  216.009 ±  0.001    B/op
LibsodiumBench.cryptoSignDetached:gc.count                                         thrpt   15    6.000           counts
LibsodiumBench.cryptoSignDetached:gc.time                                          thrpt   15    9.000               ms
LibsodiumBench.cryptoSignDetachedNoChecks                                          thrpt   15   74.370 ±  0.286  ops/ms
LibsodiumBench.cryptoSignDetachedNoChecks:gc.alloc.rate                            thrpt   15   15.320 ±  0.059  MB/sec
LibsodiumBench.cryptoSignDetachedNoChecks:gc.alloc.rate.norm                       thrpt   15  216.009 ±  0.001    B/op
LibsodiumBench.cryptoSignDetachedNoChecks:gc.count                                 thrpt   15    6.000           counts
LibsodiumBench.cryptoSignDetachedNoChecks:gc.time                                  thrpt   15    9.000               ms
LibsodiumBench.cryptoSignDetachedNoChecks_MemorySegments                           thrpt   15   74.587 ±  0.361  ops/ms
LibsodiumBench.cryptoSignDetachedNoChecks_MemorySegments:gc.alloc.rate             thrpt   15    0.001 ±  0.001  MB/sec
LibsodiumBench.cryptoSignDetachedNoChecks_MemorySegments:gc.alloc.rate.norm        thrpt   15    0.009 ±  0.001    B/op
LibsodiumBench.cryptoSignDetachedNoChecks_MemorySegments:gc.count                  thrpt   15      ≈ 0           counts
LibsodiumBench.cryptoSignDetached_MemorySegments                                   thrpt   15   74.565 ±  0.310  ops/ms
LibsodiumBench.cryptoSignDetached_MemorySegments:gc.alloc.rate                     thrpt   15    0.001 ±  0.001  MB/sec
LibsodiumBench.cryptoSignDetached_MemorySegments:gc.alloc.rate.norm                thrpt   15    0.009 ±  0.001    B/op
LibsodiumBench.cryptoSignDetached_MemorySegments:gc.count                          thrpt   15      ≈ 0           counts
LibsodiumBench.cryptoSignVerifyDetached                                            thrpt   15   26.686 ±  0.177  ops/ms
LibsodiumBench.cryptoSignVerifyDetached:gc.alloc.rate                              thrpt   15    5.498 ±  0.037  MB/sec
LibsodiumBench.cryptoSignVerifyDetached:gc.alloc.rate.norm                         thrpt   15  216.026 ±  0.001    B/op
LibsodiumBench.cryptoSignVerifyDetached:gc.count                                   thrpt   15      ≈ 0           counts
LibsodiumBench.cryptoSignVerifyDetachedNoChecks                                    thrpt   15   26.479 ±  0.125  ops/ms
LibsodiumBench.cryptoSignVerifyDetachedNoChecks:gc.alloc.rate                      thrpt   15    5.455 ±  0.026  MB/sec
LibsodiumBench.cryptoSignVerifyDetachedNoChecks:gc.alloc.rate.norm                 thrpt   15  216.026 ±  0.001    B/op
LibsodiumBench.cryptoSignVerifyDetachedNoChecks:gc.count                           thrpt   15      ≈ 0           counts
LibsodiumBench.cryptoSignVerifyDetachedNoChecks_MemorySegments                     thrpt   15   26.919 ±  0.175  ops/ms
LibsodiumBench.cryptoSignVerifyDetachedNoChecks_MemorySegments:gc.alloc.rate       thrpt   15    0.001 ±  0.001  MB/sec
LibsodiumBench.cryptoSignVerifyDetachedNoChecks_MemorySegments:gc.alloc.rate.norm  thrpt   15    0.025 ±  0.001    B/op
LibsodiumBench.cryptoSignVerifyDetachedNoChecks_MemorySegments:gc.count            thrpt   15      ≈ 0           counts
LibsodiumBench.cryptoSignVerifyDetached_MemorySegments                             thrpt   15   26.777 ±  0.126  ops/ms
LibsodiumBench.cryptoSignVerifyDetached_MemorySegments:gc.alloc.rate               thrpt   15    0.001 ±  0.001  MB/sec
LibsodiumBench.cryptoSignVerifyDetached_MemorySegments:gc.alloc.rate.norm          thrpt   15    0.026 ±  0.001    B/op
LibsodiumBench.cryptoSignVerifyDetached_MemorySegments:gc.count                    thrpt   15      ≈ 0           counts
```

abies:
Please note that for us, speed of verification is a LOT more important than signing speed. You need to verify everything coming from all the other nodes and users, while sign only your own stuff. So, if one library has zero memory pressure for verification, it might be better, even if it has slightly worse ops/s.

anthony-swirldslabs:
@abies : 
> Please note that for us, speed of verification is a LOT more important than signing speed. You need to verify everything coming from all the other nodes and users, while sign only your own stuff. So, if one library has zero memory pressure for verification, it might be better, even if it has slightly worse ops/s.

Understood. And to that end, we'll be caching some of the MemorySegments to avoid (re-)allocating the same objects. In particular, PBJ will automatically cache results of Bytes.toMemorySegment(). Also, SodiumVerifier will cache a MemorySegment for its publicKey. All this caching, together with an improved TPS, should benefit the overall performance of the Consensus Node.

anthony-swirldslabs:
@abies : There's one more point to consider - the existing `SodiumVerifier` calls `Bytes.toByteArray()` on the signature and message inputs. This operation allocates a new array and copies the original data into it. So I went ahead and update the benchmark to verify this case (see the latest commit), and the results are below. As you can see, the `cryptoSignVerifyDetached_Bytes` variant has `gc.alloc.rate.norm` at around 160 B/op. So the old number of zero was misleading when the Consensus Node implementation is taken into account, and the new number is not too far from the non-MemorySegment-caching version of the FFM API that we built. So I don't see a reason for a concern on this part.

```
Benchmark                                                           Mode  Cnt     Score    Error   Units
LazysodiumBench.cryptoSignDetached                                 thrpt   15    68.146 ±  1.038  ops/ms
LazysodiumBench.cryptoSignDetached:gc.alloc.rate                   thrpt   15    15.649 ±  2.397  MB/sec
LazysodiumBench.cryptoSignDetached:gc.alloc.rate.norm              thrpt   15   241.085 ± 39.203    B/op
LazysodiumBench.cryptoSignDetached:gc.count                        thrpt   15    29.000           counts
LazysodiumBench.cryptoSignDetached:gc.time                         thrpt   15  1262.000               ms
LazysodiumBench.cryptoSignVerifyDetached                           thrpt   15    25.550 ±  0.153  ops/ms
LazysodiumBench.cryptoSignVerifyDetached:gc.alloc.rate             thrpt   15     0.001 ±  0.001  MB/sec
LazysodiumBench.cryptoSignVerifyDetached:gc.alloc.rate.norm        thrpt   15     0.027 ±  0.001    B/op
LazysodiumBench.cryptoSignVerifyDetached:gc.count                  thrpt   15       ≈ 0           counts
LazysodiumBench.cryptoSignVerifyDetached_Bytes                     thrpt   15    25.715 ±  0.233  ops/ms
LazysodiumBench.cryptoSignVerifyDetached_Bytes:gc.alloc.rate       thrpt   15     3.924 ±  0.035  MB/sec
LazysodiumBench.cryptoSignVerifyDetached_Bytes:gc.alloc.rate.norm  thrpt   15   160.028 ±  0.003    B/op
LazysodiumBench.cryptoSignVerifyDetached_Bytes:gc.count            thrpt   15       ≈ 0           counts
```

Jeffrey-morgan34 (approved):

